PR #23, opened 2015-03-10
Fix Tokenizer.hasNext()

danialgoodwin:
Previously Tokenizer.hasNext() would return true when there was only
trailing whitespace. It now correctly returns false in that case and when
input is only whitespace


squeek502:
:+1:

uklimaschewski:
Thank you very much!

PR #31, opened 2015-07-03
Please make junit a test-scoped dependency

FauxFaux:
This stops the dependency leaking into other projects when they depend on the library.

It also helps stop you using JUnit classes in production code.

You can read more about test scopes in the Maven documentation:
https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope


uklimaschewski:
Thank you! I *love* maven ... not. ;-)

PR #42, opened 2015-09-05
Small performance enhancement

bbarnhill:
use non-synchronized string builder for better performance since locks arent blocking the thread, use Locale.getDefault instead of Locale.ROOT

### DIFF
--- a/src/com/udojava/evalex/Expression.java
+++ b/src/com/udojava/evalex/Expression.java
@@ -279,7 +279,7 @@ public class Expression {
 		 *            The number of parameters for this function.
 		 */
 		public Function(String name, int numParams) {
-			this.name = name.toUpperCase(Locale.ROOT);
+			this.name = name.toUpperCase(Locale.getDefault());
 			this.numParams = numParams;
 		}
 
@@ -857,7 +857,7 @@ public class Expression {
 				outputQueue.add(token);
 			} else if (variables.containsKey(token)) {
 				outputQueue.add(token);
-			} else if (functions.containsKey(token.toUpperCase(Locale.ROOT))) {
+			} else if (functions.containsKey(token.toUpperCase(Locale.getDefault()))) {
 				stack.push(token);
 				lastFunction = token;
 			} else if (Character.isLetter(token.charAt(0))) {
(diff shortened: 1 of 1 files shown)


squeek502:
Locale.getDefault() has the same effect as omitting the Locale parameter completely, which poses problems in different languages.

See these blog posts:
 * http://www.moserware.com/2008/02/does-your-code-pass-turkey-test.html
 * http://blog.codinghorror.com/whats-wrong-with-turkey/

bbarnhill:
Explain to me how Locale.ROOT is better since it doesnt specify a country or language and getDefault atleast attempts to get the users locale.  I just dont understand where Locale.ROOT is better.

squeek502:
Because in certain Locales, Turkish for instance, toUpperCase and toLowerCase behave in ways that you might not expect. The blog posts I linked explain it in detail, but basically, if you do:

```java
String test = "i use the letter i";
System.out.println(test.toUpperCase(), new Locale("tr", "TR"));
```
you'd get:
```
İ USE THE LETTER İ
```

Note that the İ is not the uppercase I, but rather U+0130 “Latin Capital Letter I With Dot Above”. Therefore, comparisons might fail when they should really pass.

Using Locale.ROOT just makes sure that ascii characters will always remain as ascii characters when they are sent through toUpperCase()/toLowerCase(). Always forcing it to use en-US would fix the problem as well, but Locale.ROOT is the most neutral you can get and it does the job.

Finally, here's [what FindBugs has to say about it](http://findbugs.sourceforge.net/bugDescriptions.html#DM_CONVERT_CASE):

>  A String is being converted to upper or lowercase, using the platform's default encoding. This may result in improper conversions when used with international characters. Use the
```
>    String.toUpperCase( Locale l )
>    String.toLowerCase( Locale l )
```
> versions instead.

bbarnhill:
K I'll get this pull updated later today to remove those but I think the
string builder still applies.

Brad Barnhill
bradbarnhill@hotmail.com

bbarnhill:
repulled in https://github.com/uklimaschewski/EvalEx/pull/45

PR #47, opened 2015-10-16
Validate the rpn expression before eval

### LINKED ISSUE #41: Invalid expressions give results
Invalid expressions give results, e.g.:  “12 18 2” = 2. or “(12)(18)” = 18.
Instead, an exception should be thrown.

luis451:
Fixed #41

### DIFF
--- a/src/com/udojava/evalex/Expression.java
+++ b/src/com/udojava/evalex/Expression.java
@@ -1105,6 +1105,8 @@ public class Expression {
 	 */
 	public BigDecimal eval() {
 
+		validate();
+
 		Stack<BigDecimal> stack = new Stack<BigDecimal>();
 
 		for (String token : getRPN()) {
--- a/src/com/udojava/evalex/Expression.java
+++ b/src/com/udojava/evalex/Expression.java
@@ -1105,8 +1105,6 @@ public class Expression {
 	 */
 	public BigDecimal eval() {
 
-		validate();
-
 		Stack<BigDecimal> stack = new Stack<BigDecimal>();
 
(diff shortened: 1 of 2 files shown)


squeek502:
It would probably be better to move the `validate()` call to either the constructor or after the `shuntingYard()` call in `getRPN()`. Calling `validate` each `eval` is overkill since the RPN is constant per `Expression` instance, and it would cut into the benefits of #3.

uklimaschewski:
First of all, thank you luis for this code - looks like an easy and quick validation feature.
But I agree with Ryan that this should not be called on each evaluation.
You can place it here:
````java
	private List<String> getRPN() {
		if (rpn == null) {
			rpn = shuntingYard(this.expression);
			validate(rpn); // only called once for each expression
		}
		return rpn;
	}
````

luis451:
Updated with those changes.
 I moved `validate()` after the `shuntingYard()` call in `getRPN()`

uklimaschewski:
This may sound like nitpicking, but I am unhappy with the getRPN() call in validate(), this looks odd, because validate() is called from getRPN().
Could you just pass the rpn instance as a parameter to the validate() method and use it there in the for-loop?
This will look much cleaner.

luis451:
I added parameter to `validate()`, also i make `validate()` private since `getRPN()` is private and is te only way to get the rpn notation. As well being private there a no reason to `validate()` to return  `this`

uklimaschewski:
Ok, thank you again :+1:

squeek502:
Very cool. :+1:

PR #49, opened 2015-10-30
Add support for functions with variable numbers of parameters

### LINKED ISSUE #48: Max function for more than 2 arguments
```max(3.0, 2.0, 1.0)``` returns ```2.0```. The expected answer would be 3.0.
```min(1.0, 2.0, 3.0)``` returns ```2.0```. The expected answer would be 1.0.

Currently max expects two arguments. If more arguments are passed to the function, the last two are taken (no error). This is not obvious.

Could max (and min) be extended to accept any number of arguments (>1)?

squeek502:
* Implementation based on "David's method" outlined in Gene Pavlovsky's comment here: http://www.kallisti.net.nz/blog/2008/02/extension-to-the-shunting-yard-algorithm-to-allow-variable-numbers-of-arguments-to-functions/#comment-125789
 * Modifies the RPN representation by adding a ( token to denote the start of a function's parameter list
 * Made MIN and MAX take a variable number of parameters
 * Closes #48 

---

Needs review for sure. I mostly just wanted to see if I could get it to work and didn't really make sure I was implementing everything in the most proper way. Let me know if there's anything that can/should be improved. One thing I think can definitely be improved is the wording of the exceptions that were added.

Note this also has the side benefit of providing better feedback about unexpected numbers of function parameters, as it can now be known how many parameters were given to a specific function and that can be compared to the expected number.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,8 +92,8 @@ To include it in your Maven project, refer to the artifact in your pom:
   <tr><td>NOT(<i>expression</i>)</td><td>Boolean negation, 1 (means true) if the expression is not zero</td></tr>
   <tr><td>IF(<i>condition</i>,<i>value_if_true</i>,<i>value_if_false</i>)</td><td>Returns one value if the condition evaluates to true or the other if it evaluates to false</td></tr>
   <tr><td>RANDOM()</td><td>Produces a random number between 0 and 1</td></tr>
-  <tr><td>MIN(<i>e1</i>,<i>e2</i>)</td><td>Returns the smaller of both expressions</td></tr>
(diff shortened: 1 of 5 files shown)


uklimaschewski:
Nice, I'll hope I find some time this weekend to have a detailed look at it. Maybe we need some more unit tests to make sure nothing is broken?

uklimaschewski:
Thank you, Ryan! I see no incompatibility or other problems. I have added some more tests and they all work. I'll check that in later along with updated documentation.
Thanks again.

squeek502:
No problem! Was fun figuring it out.

Issue #51, opened 2015-11-05
Scientific notation


qwert2003:
Scientific notation (```"2.2E-16"```) within expressions currently is not supported.

Example (calling EvalEx from BeanShell) ...
```
addClassPath("drawj2d.jar"); // BeanShell (bsh2) syntax
import com.udojava.evalex.*;
import java.math.BigDecimal;

String expr2 = "2.2E-16 * 10.2";
Expression expression = new Expression(expr2);
BigDecimal result = expression.eval();
```
... fails throwing an exception:
```
java.lang.RuntimeException: Unknown operator or function: E
        at com.udojava.evalex.Expression.shuntingYard(Expression.java:1149)
        at com.udojava.evalex.Expression.getRPN(Expression.java:1353)
        at com.udojava.evalex.Expression.eval(Expression.java:1166)
```

I once asked in the comments section of Udo's blog. The answer was:
> Scientific notation, this could get tricky. But I guess it is possible to implement in the tokenizer.

The java method call ```new BigDecimal("2.2E-16")``` does parse numbers in scientific notation. So it seems that it would be enough to tell the EvalEx tokenizer to understand the number in scientific notation as a single statement.
Trying ...
```
Index: com/udojava/evalex/Expression.java
===================================================================
--- com/udojava/evalex/Expression.java  (revision 263)
+++ com/udojava/evalex/Expression.java  (working copy)
@@ -593,7 +593,8 @@
                                ch = input.charAt(++pos);
                        }
                        if (Character.isDigit(ch)) {
-                               while ((Character.isDigit(ch) || ch == decimalSeparator)
+                               while ((Character.isDigit(ch) || ch == decimalSeparator
+                                                || ch == 'e' || ch == 'E')
                                                && (pos < input.length())) {
                                        token.append(input.charAt(pos++));
                                        ch = pos == input.length() ? 0 : input.charAt(pos);
@@ -1056,7 +1057,8 @@
                        return false;
                for (char ch : st.toCharArray()) {
                        if (!Character.isDigit(ch) && ch != minusSign
-                                       && ch != decimalSeparator)
+                                       && ch != decimalSeparator
+                                        && ch != 'e' && ch != 'E')
                                return false;
                }
                return true;
@@ -1398,6 +1400,7 @@
                        }
                        counter++;
                }
+                System.err.println("DEBUG: " + rpn.toString());
                if (counter > 1) {
                        throw new ExpressionException("Too many numbers or variables");
[...]

PR #52, opened 2015-11-08
Numbers using scientific notation #51

qwert2003:
Support scientific notation. See #51.

The tokenizer is changed, such that it accepts numbers containing 'e'. The actual parsing is done by the java method call new BigDecimal(numberstring), which accepts numbers in scientific notation.


uklimaschewski:
Can you please provide some unit tests that show that it works? There are several possible notations, e.g.:

1e10
1e+10
1e-10
1E10
1E+10
1E-10
123.456E3
2135E-4

and possibly some more in combination with expresions that use mixed notifications with + and - operations, braces etc.

qwert2003:
For scientific notation I have added the junit test class TestSciNotation.

The existing unit tests still pass. Of course I'd appreciate if others did some more tests to verify that the code changes do not break anything.

uklimaschewski:
Thank you, looks good.

PR #55, opened 2016-02-07
add finals, general code cleanup

barnhill:
Adding finals to variables and places that it can be to more clearly denote what should be modified and what shouldnt.  Also allows for more compilation optimization.

Some general code cleanup as suggested by intelliJ to prevent coding errors in the future (ex. methods and if statement blocks that have no brackets)

### DIFF
--- a/src/com/udojava/evalex/Expression.java
+++ b/src/com/udojava/evalex/Expression.java
@@ -307,7 +307,7 @@ import java.util.Stack;
  * 
  * e.addOperator(e.new Operator(">>", 30, true) {
  *     {@literal @}Override
- *     public BigDecimal eval(BigDecimal v1, BigDecimal v2) {
+ *     public BigDecimal eval(final BigDecimal v1, final BigDecimal v2) {
  *         return v1.movePointRight(v2.toBigInteger().intValue());
  *     }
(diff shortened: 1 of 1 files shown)


uklimaschewski:
Thank you, but after a quick test, I see no performance enhancements resulting from the extensive use of the `final` keywword in the code.

Using `final` in method parameters is also of very limited use and practically not needed here. (IMHO)

Pull requests should either enhance the performance or functionality of this library.

barnhill:
http://www.javapractices.com/topic/TopicAction.do?Id=23

barnhill:
http://stackoverflow.com/questions/137868/using-final-modifier-whenever-applicable-in-java

uklimaschewski:
As I noted, this is all about coding practices. And I don't want to start a war on what is the best coding practice.
It has no influence on performance or functionality.

barnhill:
It was less about performance and more about clarity on those variables and their intention of not being set to another instance.  Its fine it works as is.  I was just going for cleaner, more concise and clear coding intentions on the variable declarations.  Its your lib and you make the ultimate decision.  Just trying to help out.  Keep up the good work on it.

PR #58, opened 2016-02-28
Expose declared variable, function as well as operators

fobbyal:
Sorry for the bad PR before.. 
This PR adds the functionality of exposing defined variables, operators as well as functions. I am working on a project that need this functionality for resolving multiple expressions have inter dependencies.
I hope the format is okay. Let me know if there are more problems

PR #59, opened 2016-02-28
Making declared functions,variables as well as operators case insensitive

fobbyal:
Sorry for the bad PR before.. 
This PR adds the feature of using case insensitive variables.

Let me know if the format is okay.

Thanks so much for the work you have done here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 /.settings
 .classpath
 .project
+*.iml
+.idea
+target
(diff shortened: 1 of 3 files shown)


uklimaschewski:
Sorry for the delay!
I have committed a corrected version of this pull request with https://github.com/uklimaschewski/EvalEx/commit/34993afaa7147de41600a6a27412867f83e23e2d.
There were two problems with your original pull requests #58 and #59:
Some import statements were missing and the new tests were not added to [AllTests.java](https://github.com/uklimaschewski/EvalEx/blob/34993afaa7147de41600a6a27412867f83e23e2d/tests/com/udojava/evalex/AllTests.java#L11).

Issue #62, opened 2016-03-30
Quick Question about BigDecimal and If function


fobbyal:
The project I am using EvalEx on is money based therefore all calculation falls in the limits of double. BigDecimal arithmetic function is known to be much more slower than primitive like double. I am exploring the possibility of wrapping the Values EvalEX in a pseudo class that supports BigDecimal as well as primitives like double or even floats in extreme case. For use cases that are can tolerate precision losses based on performance gain.

This is going to dramatically change the code so i am asking to see if you will be willing to accept such a pull request.

If i go ahead with this modification. I would like to improve the if function as well. Consider the following
`if(a =0, 0 , 12/a)`
The above does not work well because 12/a is evaluated regardless of the outcome of `a=0`. This was a problem i ran into. I solved it by returning 0 for any divisible by 0 in my specific context. This is not ideal. I would like to lazily evaluate the parameters of the if based on the outcome of the boolean tester.

uklimaschewski:
If you are working with money, `BigDecimal` should be the way to go. Unless you really need the performance of `double`. With `double`, you have to take care for proper rounding in each part of the expression, else you will end up with missing cents everywhere.
That was one of the original reasons, why I created this expression parser. All other parsers around were using `double` and were therefore unusable in a financial environment, where I needed it.

Have a look at this:
````java
    System.out.println( "362.2 - 362.6 = " + ( 362.2 - 362.6 ) );
````
it will print
````
    362.2 - 362.6 = -0.4000000000000341
````
Which is horrific, if you use this as part of a sum in a financial environment without rounding even the substraction.
You will probably not notice any problem, until the sum of the resulting doubles suddenly end up being a cent to high or low.

If you need the performance, `double` is much faster, but it will only be noticeable when calculating millions of expressions.

have a look at [this blog article](http://java-performance.info/bigdecimal-vs-double-in-financial-calculations/), which sums up the pros and cons of `BigDecimal` vs. `double` in monetary calculations.

Somebody did such a modification to EvalEx a while back, but it ended up in a more complex library than I wanted to maintain. I would like to keep this small and compact, within only one class.

The problem with the `IF()` function is a constant source of misunderstanding. Let me see how you plan to 'fix' this issue, I would be happy to include any good solution.

fobbyal:
Here is my plan for the LazyIf..
[...]

PR #63, opened 2016-04-05
lazyIf implementation

fobbyal:
Finally had sometime to implement the feature without breaking the project format. So here it is. All tests are passing including the divide by 0 scenario. Let me know if you think there could be a problem with this.


uklimaschewski:
Thank you, I'll give this a try. See my comments in issue #62

PR #68, opened 2016-05-08
add the Definition of e: "Euler's number" as a constant.

humazed:
I think `e` must be part of the constants so I added it.


uklimaschewski:
Thank you, I merged this.

PR #71, opened 2016-06-12
Fix empty stack errors and improve error handling for operators

squeek502:
- An operator at the end of a function parameter list (`min(3/)`) would throw an EmptyStackException
 - Made all correctness checking done in validate() rather than some being done in eval()
 - shuntingYard() now catches operators that don't have enough parameters at the end of function parameter lists that would otherwise provide a valid RPN (`min(1,2,3/)` would give the valid RPN `( 1 2 3 / min`, but the expression syntax is actually invalid)
 - shuntingYard() now catches operators that don't have enough parameters in the middle of function parameter lists that would otherwise provide a valid RPN (`min(1,2/,3)` would give the valid RPN `( 1 2 / 3 min`, but the expression syntax is actually invalid)
 - shuntingYard() now catches operators that don't have enough parameters in the middle of function parameter lists (with the operator directly following a comma/open paren) that would otherwise provide a valid RPN (`min(1,2,/3)` would give the valid RPN `( 1 2 / 3 min`, but the expression syntax is actually invalid)

### DIFF
--- a/src/com/udojava/evalex/Expression.java
+++ b/src/com/udojava/evalex/Expression.java
@@ -1133,6 +1133,10 @@ public class Expression {
 							+ lastFunction + "'");
 				}
 			} else if (operators.containsKey(token)) {
+				if (",".equals(previousToken) || "(".equals(previousToken)) {
+					throw new ExpressionException("Missing parameter(s) for operator " + token +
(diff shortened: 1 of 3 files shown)


uklimaschewski:
Thanks alot!

PR #73, opened 2016-06-16
Update README.md

zurche:
Hello, I had to add this to an Android Application I'm working on and discovered that the dependency import is only in the example for maven projects, just adding this in case someone does not know that they can use a maven dependency in gradle if it is stored in mavencentral. 

Regards,
Alejandro


uklimaschewski:
Thank you for this update!

uklimaschewski:
I changed your gradle example to gradle syntax in markup.

PR #81, opened 2016-08-31
Implement equals(), hashCode() and toString()

gangeli:


uklimaschewski:
Sorry for the late answer. You are using the Java-7 `Objects` class in your pull request. This would require moving the project from Java-6 to Java-7. I am using EvalEx in a large project, which is based on Java-6.
Do you see any chance to stay with Java-6?

gangeli:
Sure, it's fixed so that it doesn't use the Objects class.

uklimaschewski:
I merged this pr and made one change to it:
- renamed function `originalExpression()` to `getOriginalExpression()`

PR #87, opened 2016-11-28
Fixes #80: Make Operator and Function classes static

### LINKED ISSUE #80: Make Function and Operator static
If made static, I wouldn't need to create a `Function` dependent of the variable, the same way you do inside Expression class. 

There should be no other changes to the code Also, I could use it in Kotlin ;D

shartte:
By making the inner classes static, custom functions can now be defined outside of the context of an Expression object.

For example, one could create subclasses of Function as top-level classes that are singletons, and expose these to Expressions.

Additional enhancements later may allow for Java8 method references to be used directly as Functions.

### DIFF
--- a/src/com/udojava/evalex/Expression.java
+++ b/src/com/udojava/evalex/Expression.java
@@ -451,7 +451,7 @@ public class Expression {
 		BigDecimal eval();
 	}
 
-	public abstract class LazyFunction {
+	public static abstract class LazyFunction {
 		/**
 		 * Name of this function.
 		 */
@@ -497,7 +497,7 @@ public class Expression {
 	 * defined by a name, the number of parameters and the actual processing
 	 * implementation.
 	 */
-	public abstract class Function extends LazyFunction {
+	public static abstract class Function extends LazyFunction {
 
 		public Function(String name, int numParams) {
 			super(name, numParams);
--- a/src/com/udojava/evalex/Expression.java
+++ b/src/com/udojava/evalex/Expression.java
@@ -530,7 +530,7 @@ public class Expression {
 	 * Abstract definition of a supported operator. An operator is defined by
 	 * its name (pattern), precedence and if it is left- or right associative.
 	 */
-	public abstract class Operator {
+	public static abstract class Operator {
 		/**
 		 * This operators name (pattern).
 		 */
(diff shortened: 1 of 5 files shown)


uklimaschewski:
I'm having a problem with the fact that your changes are incompatible with existing implementations that use custom functions and/or operators. You had to rewrite all the unit tests, that use custom functions or operators and all existing implementations out there will have to do that with their custom functions/operators, too. 
Can you think of a way to get what you want without breaking compatibility?

shartte:
Yeah I can absolutly see your point. Making such a non-backwards compatible change should be well thought out.

Since the PR i actually discovered a convenient way to make custom functions without this change, which involves making a custom subclass of Expression and registering all custom functions in the constructor.

The only thing I can think of that would make this change backwards compatible is leaving the old Function/Operator classes around, and introducing a new separate Function/Operator classes that would be static or maybe even no longer an inner class.

How much are you hanging onto the fact that your library is effectively just a single java file?

brielmayer:
There is no way for this kind of change not to break compatibility. In my opinion you shouldn't care because you are not able guarantee full backward compatibility through the whole development process. If someone wants to use the new version, they have to adapt their code. Easy as that...

uklimaschewski:
@brielmayer I have created EvalEx as part of a large commercial software product, where it is used by >100 customers. A change would, at least, require our customers to rebuild their code. (It is a some kind of programming system).
This project has also 96 forks and therefore probably alot of people using it in their products, too.
Breaking backward compatibility should always be the last possibility to consider.
Without a real important reason, I will not do it.
But I will think about some possibilities to go this way without loosing backward compatibility.

uklimaschewski:
@shartte @brielmayer and others:
I have checked in a new branch: https://github.com/uklimaschewski/EvalEx/tree/Externalize
Here I added two external classes `ExternalFunction` and `ExternalLazyFunction`.
They can be added through `Expression.addExternalFunction(...)`.
See the test `TestExternalFunctions`.

uklimaschewski:
I'm closing this PR for now, the topic is still on the table and I will possibly move the code into my proposed direction.

gredler:
I was also surprised by this design decision today -- it was the first issue that I ran into in the first hour of evaluating this library.

PR #89, opened 2017-01-06
Allow legal variable characters to be modified

rjcohn:
Added 2 fields (and setters) to Expression, firstVarCharacters and varCharacters that specify the legal characters for the first character and subsequent characters of a variable in addition to letters and numbers. By default, the only legal character for both is "_", the same as before.
Added tests that use both setters.


squeek502:
Looks like e6e5b27 reverted your test class to a copy of TestEval.

uklimaschewski:
Thank you. I have merged it and created a new release 1.5. Should be public on Maven Central in a short while.

rjcohn:
Thanks! That was quick.

PR #91, opened 2017-01-16
Fix 2 bugs in firstVarChars

rjcohn:
1. Handle when an operator is adjacent to a char in firstVarChars that's not in the default set.
2. Handle when a var starts with a char in firstVarChars that's not in varChars.
See testFirstVarChar for both.

Sorry for missing these before.


uklimaschewski:
thanks, I'll trigger a release build

PR #97, opened 2017-03-04
Separate into files

iamsubhranil:
@uklimaschewski while using with IntelliJ Idea, I suddenly gotten in 'class not enclosed error'. I jumped into the source and found a plethora of enclosing classes, and thought to clean up the project by separating them into files. It certainly increases readability by quite a few degrees, and helps to manage the core Expression code in a more efficient way. How about taking a glimpse at it?
P.S. I have not updated the test suite with the new classes.
P.S. 2. : I've added FACT and EXP functions


uklimaschewski:
I'm closing this for now, see my last comment to PR #80. I will possibly move the code into my proposed direction of separate files.

PR #100, opened 2017-06-02
update maven dependency version in readme

ugurk:
Old one may cause wrong usage for inattentive developers


uklimaschewski:
Thanks.

PR #102, opened 2017-06-13
Support null as value

rbuehlma:
With this change, NULL values are supported. If an operation does not support "null" as value, it will throw an exception.

NULL values can be used for functions like COALESCE() in SQL where the first non-null value is used.


uklimaschewski:
I have merged PR #99 from @revbingo which changed the tokenizer quite substantly (to the better, IMHO).
As a result your (and other) PRs now have conflicts.
Please have a look at the new code, if you want you can apply your changes again and open a new PR.
Until then, I'm closing this one.

Issue #84, opened 2016-11-15
factorial operator


nietoperz809:
insert the following in the operators list:

        ///////////////////////////////////////////////////
        addOperator(new Operator("!", 50, true)
        {
            BigDecimal fac(BigDecimal n, BigDecimal acc)
            {
                if (n.equals(BigDecimal.ONE))
                {
                    return acc;
                }
                BigDecimal lessOne = n.subtract(BigDecimal.ONE);
                return fac(lessOne, acc.multiply(lessOne, mc));
            }

            @Override
            public BigDecimal eval (BigDecimal v1, BigDecimal v2)
            {
                return fac (v1, v1);
            }
        });
        ///////////////////////////////////////////////////

... then replace every ! in the input string with !0 to fake a second op
now one can calculate 
3!^4 = 1296
3! = 6
8! = 40320
and so on ...

squeek502:
Why not make it a function instead?

nietoperz809:
for the same reason why + is not a function: x = add (a,b)
! is the usual mathematical notation. :)

squeek502:
Right, but there's currently no support for unary operators (see #36), so making it a function would remove the need for the search and replace hack.

nietoperz809:
it's not a hack, it's a nice solution. :)
inserting fake operands enables this software to deal with unary operators. i think it will work in aller other cases (like negation, increment, etc..)

uklimaschewski:
I agree with Ryan, a general support for unary operators would be better.
Have a look at issue #36, user yoavst has published some code, which is a good starting point.
Would be happy to have this feature in EvalEx.

PR #110, opened 2017-07-16
Unary operators

### LINKED ISSUE #36: Support for Unary operators
So we could write `+5` or `-SQRT(9)`
Currently it throws exception.

revbingo:
This resolves #36, and I claim my eternal fame and honour :)

Only unary plus and minus supported at this point, factorial (issue #84) is a little trickier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,8 @@ dependencies {
 <table>
   <tr><th>Mathematical Operators</th></tr>
   <tr><th>Operator</th><th>Description</th></tr>
-  <tr><td>+</td><td>Additive operator</td></tr>
-  <tr><td>-</td><td>Subtraction operator</td></tr>
+  <tr><td>+</td><td>Additive operator / Unary plus</td></tr>
+  <tr><td>-</td><td>Subtraction operator / Unary minus</td></tr>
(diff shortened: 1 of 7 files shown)


uklimaschewski:
Thank you, but I found a problem. I added some a bit more complex tests and they failed.
E.g.:
		assertEquals("-2.75", new Expression("-2+3/4*-1").eval().toPlainString());
		assertEquals("-1.25", new Expression("-2+3/4").eval().toPlainString());

Both fail with a wrong result.
Can you have a look at your logic and also add a few more tests for unary operators?

revbingo:
Apologies, bang to rights.  There were actually two problems, the shuntingYard algorithm needed to check additionally for UNARY_OPERATOR in the while loop, and also it was getting the operator to compare against from the operators map, which didn't include unaryOperators.

I've fixed this by putting unaryOperators in the operators map too, but with a "u" suffix to disambiguate them, and therefore also adding a `u` to `token.surface` when detected. This or similar seems to be a relatively common solution to the problem.

uklimaschewski:
OK, looks good so far. I'll make a new release tomorrow. I want to update the README with some of the features that were added in the last months.

PR #118, opened 2017-08-10
atan2

qwert2003:
Add ATAN2 function.


uklimaschewski:
Can you add documentation (Readme.md and JavaDoc) and also at least one unit test for the new function? Thank you.

PR #137, opened 2017-10-31
Make numbers starting with the decimal separator match (such as ".1")

baumeister:


uklimaschewski:
Sorry, but this change breaks several unit tests. Can you check this and also provice some tests for the new feature?

baumeister:
Done.

PR #143, opened 2017-12-12
Added support for nested variables

nictomlinson:
This is the pull request related to the enhancement we discussed today. As noted, it tries to deal with cases such as this:
```Java
e = new Expression("x+y")
e.with("x","a+1") // e is now "a+1+y"
e.with("y","x+a") // e is now "a+1+x+a"
e.with("a",2) // e is now "2+1+x+2"
e.eval() // even though we have already defined x, this will fail
```
So that it does not matter what order variables are defined and does not matter at what depth or how frequently a variable is referenced in either the expression or other variables.

The current pull request has a defect in that it does not handle user defined operators but this should be easy to fix in a similar style to user defined functions. For now, it is the concept I'd like to discuss.

There are a couple of test cases added to TestNested.java.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,7 @@
 		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
-			<version>4.4</version>
+			<version>4.12</version>
 			<scope>test</scope>
 		</dependency>
 	</dependencies>
--- a/src/main/java/com/udojava/evalex/Expression.java
+++ b/src/main/java/com/udojava/evalex/Expression.java
@@ -503,7 +503,8 @@ public class Expression {
 	/**
 	 * All defined variables with name and value.
 	 */
-	private Map<String, BigDecimal> variables = new TreeMap<String, BigDecimal>(String.CASE_INSENSITIVE_ORDER);
+	private Map<String, LazyNumber> variables = new TreeMap<String, LazyNumber>(String.CASE_INSENSITIVE_ORDER);
+
 
 	/**
 	 * What character to use for decimal separators.
@@ -547,6 +548,23 @@ public class Expression {
 	public interface LazyNumber {
 		BigDecimal eval();
 		String getString();
+	}
+
+	/**
+	 * Construct a LazyNumber from a BigDecimal
+	 */
+	private LazyNumber LazyNumber(final BigDecimal bigDecimal){
+		return new LazyNumber(){
+				@Override
+				public String getString() {
+					return bigDecimal.toPlainString();
+				}
+		
+				@Override
+				public BigDecimal eval() {
+					return bigDecimal;
+				}
+		};	
 	}
 
 	public abstract class LazyFunction {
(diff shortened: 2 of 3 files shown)


uklimaschewski:
@nictomlinson I had in mind, that there was a similar discussion a while ago.
See the discussion in issue #9 (support for nested expressions)
And see this commit: https://github.com/uklimaschewski/EvalEx/commit/b4617c109a8ca0d34c55e99abc5ea070166c8f9f
Since 2014, things changed and we now have that lazy number feature, which might help.
I have to take a more detailed look at your changes, hopefully this weeken I will find some time.

nictomlinson:
@uklimaschewski ,
Sounds like the same discussion and the use case is similar to that of the [benze](https://github.com/uklimaschewski/EvalEx/issues/9#issuecomment-49683978).
I think my proposed change keeps the library simple to use; in fact simpler as the user does not need to worry about the order of variable definition.
I would appreciate you taking the time to look at the proposed code change.
All the current test cases pass as do the two new ones.

uklimaschewski:
I had a look at it and would like to accept this PR. But you talked about a defect with user defined operators. Can you fix this, too? Then I can merge your changes directly.
And maybe we should rename the method
```java
    private LazyNumber LazyNumber(final BigDecimal bigDecimal)
```
to
```java
    private createLazyNumber LazyNumber(final BigDecimal bigDecimal)
```
(Just to follow programming conventions)

nictomlinson:
Cheers, that good news.
Yes, I'll fix the defect with nested operators and rename the method.
I wonder if you could have a think about the other externally facing functions and how this might impact them. For example, would there be a breaking change on the toRPN method. With your latest version, if you have:
```
e = new expression("x+y");
e.with("x", "a+b");
```
what do you expect to get if you then e.toRPN and would you (do you?) get the same with the pull request.

As I don't know the library that well, I'm not sure what other aspects I need to test. If you can create or suggest some test cases, I'm happy to fix them if they fail before we commit it to master.

I'll fix the operator method name and wait to hear from you.

P.S. would you be interested in a pull request to fix issues arising from the strict javadoc handing in java8 - just a bunch of changes of special characters to the HTML encodings?

uklimaschewski:
I will wait with a new release until the externalize functionality is done by @RobertZenz.
If you find some time now, you can finish the PR. I can then merge it and Robert can start working after the holidays, as he planned, with your changes already in the code.
I will then create release 2.0 of EvalEx.
Though this indicates that there might be some incompatibilities in the internals, it would be nice to limit them to the minimum.
So, it would be nice if
```Java
	@Test
	public void testNested() {
		Expression e = new Expression("x+y");
		e.with("x", "a+b");
		e.with("a", "3");
		e.with("b", "4");
		e.with("y", "5");
		assertEquals("a b + y +", e.toRPN());
	}
```
would also pass after your changes, as this is what is expected from an RPN converter. isn't it?

uklimaschewski:
With JavaDoc enhancements: Go ahead. Though EvalEx is currently using Java 6, I hope this won't impact compiling and usage.

nictomlinson:
@uklimaschewski 
I've updated the branch to as discussed. Please take a look and let me know if you can now merge this in.

uklimaschewski:
Thanks alot, I'll check it in detail after christmas, possibly on wednesday. From a first view, it looks good.

Issue #155, opened 2018-02-06
Make LazyNumber an outer interface and add an accessor to add variable of LazyNumber type


RoiSoleil:
Make LazyNumber an outer interface and add an accessor to add variable of LazyNumber type

uklimaschewski:
Sorry for the late answer, @RobertZenz @RoiSoleil .
If you want to do this, Robert, I'd be happy. I am very busy at the moment with other projects and my job.

RobertZenz:
@uklimaschewski No worries. I'll look into it.

uklimaschewski:
In your PR, you mentioned that probably externalizing `LazyNumber` isn't necessary, because it is a public interface.
So, the following is possible, also outside of the `Expression` class:
```Java
		LazyNumber ln = new LazyNumber() {
			@Override
			public String getString() {
				return "whatever";
			}
			@Override
			public BigDecimal eval() {
				return BigDecimal.ZERO;
			}
		};
```
So, I think the only thing that is needed are two new methods:
```Java
public Expression with(String variable, LazyNumber value);
public Expression and(String variable, LazyNumber value);
```
I see it the same, especially as externalizing `LazyNumber` would break backward compatibility too much.

RoiSoleil:
It's ok for me if we only have the  second commit.

`public Expression with(String variable, LazyNumber value);
public Expression and(String variable, LazyNumber value);`

uklimaschewski:
I merged your PR, @RobertZenz . Thanks alot.

RoiSoleil:
Thx

PR #157, opened 2018-03-02
#155 Extracted LazyNumber and added setVariable(String, LazyNumber) overload.

RobertZenz:
Description
===========

This change adds overloads which accepts `LazyNumber`.

Changes
=======

 * New overload `setVariable(String, LazyNumber)`
 * New overload `with(String, LazyNumber)` (forwards to `setVariable`)
 * New overload `and(String, LazyNumber)` (forwards to `setVariable`)
 * `setVariable(String, BigDecimal)` forwards to `setVariable(String, LayzNumber)`

Backwards compatibility
=======================

These changes should be fully backwards compatible.

Copyright
=========

I waive my copyright to any of these changes.

### DIFF
--- a/src/main/java/com/udojava/evalex/Expression.java
+++ b/src/main/java/com/udojava/evalex/Expression.java
@@ -1469,7 +1469,20 @@ public class Expression {
 	 * @return The expression, allows to chain methods.
 	 */
 	public Expression setVariable(String variable, BigDecimal value) {
-		variables.put(variable, CreateLazyNumber(value));
+		return setVariable(variable, CreateLazyNumber(value));
+	}
+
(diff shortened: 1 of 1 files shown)


RobertZenz:
On second thought, I don't think extracting `LazyNumber` is necessary. As it is an interface, it is always referenced in a static manner (unlike `Expression.LazyFunction`). So that might not be needed for this.

RobertZenz:
I'v updated this pull request with only the overloads.

RobertZenz:
On another note, I just noticed that this allows to remove variables by calling `and("x", (LazyNumber)null)`, is that of any interest to you?

Issue #156, opened 2018-02-16
Setting Scope for Evaluation


nikeee:
Is it possible to set a scope containing all variables for evaluation?
Something like this:
```Java
Map<String, BigDecimal> scope = new HashMap<>(); // contains values for a and b
scope.put("a", BigDecimal.ZERO);
scope.put("b", BigDecimal.ONE);
Expression ex = new Expression("a+b");
ex.eval(scope);
```
It would be nice if this wouldn't mutate the internal state of `ex`, so the expression can be used across multiple threads.

RobertZenz:
What exactly are you trying to do, if I may ask? Using the same expression over multiple threads sounds interesting.

nikeee:
I have a set of expressions which use variables and a lot of input values. I only want to parse and optimize the expression once. After that, I'd like to use the expression instance on multiple threads. Each thread uses the same set of variables, but with different values. This is only possible if the expression class is not mutable and therefore thread safe. Also, the set of variables is known and can be set all at once, instead of setting them one by one using `with`.

RobertZenz:
I don't believe that a hard-fork would be necessary, instead the API could be extended to support this use-case. What would you think about something like this?

    Expression expression = new Expression("a+b+x");
    expression.with("a", BigDecimal.ZERO);
    
    Scope scope = new Scope();
    scope.with("b", BigDecimal.ZERO);
    scope.with("c", BigDecimal.ZERO);
    
    expression.eval(scope);

By default an `Expression` holds an "internal" scope of variables, but one can also set an additional scope for the evaluation. If a variable is defined in both, the external scope wins.

But note that I do not know whether `Expression` is actually thread-safe, that is a different topic I'm afraid.

RobertZenz:
I don't believe that a hard-fork is necessary. The extension of the API which you outline in your first example seems easy enough. The only thing I'd change is that not a `Map` is accepted, but instead a domain class, more like this:

    Expression expression = new Expression("a+b+x");
    expression.with("a", BigDecimal.ZERO);
    
    Scope scope = new Scope();
    scope.with("b", BigDecimal.ZERO);
    scope.with("c", BigDecimal.ZERO);
    
    BigDecimal result = expression.eval(scope);

But I'm not sure whether `Expression` is thread-safe or not.

nikeee:
This seems fine.

RobertZenz:
[...]

PR #158, opened 2018-03-02
WIP: #156 Preview of thread support.

RobertZenz:
This is just a preview of the upcoming thread support and is not done yet.

The goal is to provide support to use an `Expression` in multiple threads with different variables/functions/operators. For this a new class `Scope` is introduced, which holds the variables/functions/operators of each `Expression`.

### DIFF
--- a/src/main/java/com/udojava/evalex/Expression.java
+++ b/src/main/java/com/udojava/evalex/Expression.java
@@ -153,22 +153,7 @@ public class Expression {
 	 */
 	private List<Token> rpn = null;
 
-	/**
-	 * All defined operators with name and implementation.
-	 */
-	private Map<String, LazyOperator> operators = new TreeMap<String, LazyOperator>(
-			String.CASE_INSENSITIVE_ORDER);
-
-	/**
-	 * All defined functions with name and implementation.
-	 */
-	private Map<String, com.udojava.evalex.LazyFunction> functions = new TreeMap<String, com.udojava.evalex.LazyFunction>(
-			String.CASE_INSENSITIVE_ORDER);
-
-	/**
-	 * All defined variables with name and value.
-	 */
-	private Map<String, LazyNumber> variables = new TreeMap<String, LazyNumber>(String.CASE_INSENSITIVE_ORDER);
+	private Scope scope = new Scope();
 
 	/**
 	 * What character to use for decimal separators.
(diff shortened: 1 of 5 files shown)


RobertZenz:
Feedback is welcome.

I plan on adding some sort of "chaining" to scopes and I want to extract a "DefaultScope" which contains all the default operators and functions.

RobertZenz:
I believe this is as done as it gets. I mean, the basics are done. Please review this when you find the time and give me feedback on it, whether it is something you want (in one way or another), or not. I'm not sure about it, to be honest, there are a few interesting ideas in here, but whether this implementation is something that should stay, I can not say.

First, let me start with the positive things. Making `Expression` thread-safe was easy. The new `Scope` class is just a stupid container for functions, variables and operators. `Expression`s can now be evaluated by passing an additional `Scope` to them providing variables, functions and operators. This allows to use the same `Expression` for different input sets.

There are a few downsides. For example the additional `Scope` does not know anything about the `Expression`, so it cannot access the `MathContext`, for example, which means that you can't add a number as String to a `Scope`, it must be a `BigDecimal` or `LazyNumber`. The logic for retrieving numbers, functions and operators is a little bit more complicated, obviously.

RobertZenz:
I've been thinking about this further the last few days and tried to implement a similar idea which I called "context support". But the truth is, there is nothing to be gained without changing the API quite substantially...and there's nothing to gain with that either.

My main idea was to have a `Context` which allows to gather functions/operators/variables. There would be the default `Context` which contains all the default functions/operators/variables. A constructor would allow to override this with a custom one, like this:

```java
public Expression() {
    this(new DefaultContext());
)

public Expression(Context context) {
    super();

    this.context = context;
}
```

That would allow to control the available functions rather fine-tuned. But that would change the internals quite a bit. Also, it would mean that the `Context` must be changed for every instance of the `Expression`, which is a rather weird way to look at it.

---

Long story short, after a lot of thinking, I'm closing this as it adds *nothing* to the mix in a real-world scenario.

What *might* be interesting to do in the future would be to revise the API as a whole. Splitting everything up into their own classes and implementations and providing a revised API. I might be working on that for fun, but I don't expect it to make it into EvalEx in the long run. From time to time it's just fun to refactor things and see where that leads, like this.

PR #163, opened 2018-05-05
Added FACT(expression) (factorial) function

DesigningKnights:
I was in need of a factorial function for an app I'm working on, and I figured that rather than making custom functions, it would be great to have the factorial function included directly into the library. 

I first tried to add ! as a unary left associative operator, but it would not work.  The problem was the tokenizer would not accept it. If I made it right associative it would work, but the factorial operator always goes after the number (eg:  1!, 2!).  Since this would not work, I made it into a function instead.

The function is FACT, and takes only one parameter.  If you enter 0, it will return 1 as mathematically defined.  If you enter anything less than 0, it will still return 1.  Technically there is no definition for negative factorials, 

The reason for making the LazyNumber v1, and v2 final is that they are accessed from an internal class and should be final.

### DIFF
--- a/src/main/java/com/udojava/evalex/AbstractOperator.java
+++ b/src/main/java/com/udojava/evalex/AbstractOperator.java
@@ -66,14 +66,14 @@ public abstract class AbstractOperator extends AbstractLazyOperator implements O
 		super(oper, precedence, leftAssoc);
 	}
 
-	public LazyNumber eval(LazyNumber v1, LazyNumber v2) {
+	public LazyNumber eval(final LazyNumber v1, final LazyNumber v2) {
 		return new LazyNumber() {
 			public BigDecimal eval() {
-				return AbstractOperator.this.eval(v1.eval(), v2.eval());
+				return AbstractOperator.this.eval(v1.eval(),v2.eval());
 			}
 
 			public String getString() {
-				return String.valueOf(AbstractOperator.this.eval(v1.eval(), v2.eval()));
+				return String.valueOf(AbstractOperator.this.eval(v1.eval(),v2.eval()));
 			}
 		};
 	}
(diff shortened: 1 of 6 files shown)


RobertZenz:
There's a lot of noise in the diff (reformatting, whitespace, `final`, etc.). Are you aware of that, was that intentional?

DesigningKnights:
Wasn't intentional.

RobertZenz:
Can you remove those whitespace changes so that the diff becomes easier readable?

DesigningKnights:
I thought about that, and it will make another diff as it would be a new commit, unless I can edit manually without having to commit again.

RobertZenz:
You can always either amend the current commit (`git commit --amend`), squash multiple commits (`git rebase -i HEAD~2`) or create a completely new commit to replace this one. In all scenarios, you can then force push (`git push --force YOURREPO master`) to update this pull request.

DesigningKnights:
Actually, the whitespace changes were to clean the code itself to remove all the extraneous whitespace that was there before.  It makes the code itself more readable and properly indented and formatted.  after going back to look at the diff I remembered that IDEA autoreformats and cleans up code.  So the diff may not be pretty, but the code is formatted better.

RobertZenz:
Changing the whitespace/formatting of the code would be a completely different topic in my opinion, one that should be talked about in its own issue. As far as this pull request is concerned, it would be nice if the diff is as minimal as possible and only contains what it actually does. That makes the history of the repository cleaner and makes it easier to diagnose issues later one (the extreme case would be a commit which just changes three lines, but reformats the complete the file, when bisecting determined that that commit is at fault, it would be pretty hard to figure out *what* is the problem in it).

PR #165, opened 2018-05-16
changes testing for convergence for SQRT to avoid infinite loop for s…

ayanovsk:
…ome values

I was trying to run:

 import com.udojava.evalex.Expression;

import java.math.BigDecimal;

public class UsingEvalEx {

    public static void main(String[] args) {

     
        Expression ex2 = new Expression("sqrt(5.86547)");
        System.out.println("ex2 = " +ex2.eval());

        Expression ex3 = new Expression("SQRT(5.8654785)");
        System.out.println("ex3 = " + ex3.eval());

        Expression ex4 = new Expression("SQRT(5.8654784912345)");
        System.out.println("ex4 = " + ex4.eval());

        System.out.println("Math.sqrt for ex5: " + Math.sqrt(5.8654786));

        Expression ex5 = new Expression("SQRT(5.8654786)");
        System.out.println("ex5 = " + ex5.eval());

        Expression ex6 = new Expression("SQRT(5.865478515625)");
        System.out.println("ex6 = " +ex6.eval());

        

        System.exit(0);

    }

}

That is what I'm seeing:

C:\install_folder\open_source\UsingEvalEx\target\classes>java UsingEvalEx
ex2 = 2.4218732
ex3 = 2.4218748
ex4 = 2.4218748
Math.sqrt for ex5: 2.421875017419355

and then it stops, and never returns from ex5 :
Expression ex5 = new Expression("SQRT(5.8654786)");
        System.out.println("ex5 = " + ex5.eval());

if I change digit "6" in the end to "5"  like that new Expression("SQRT(5.8654785)");
then it calculates OK.

I'm using java 8 on Windows 10.
Tried it within IntellijIdea and from the command line as above.

I then changed the test at the exit from do ... while loop in the end of the SQRT section and now it works for me. Don't know if that change  is correct. Please, look at it.

### DIFF
--- a/src/main/java/com/udojava/evalex/Expression.java
+++ b/src/main/java/com/udojava/evalex/Expression.java
@@ -1,9 +1,9 @@
 /*
  * Copyright 2012-2018 Udo Klimaschewski
- * 
+ *
  * http://UdoJava.com/
  * http://about.me/udo.klimaschewski
- * 
+ *
  * Permission is hereby granted, free of charge, to any person obtaining
  * a copy of this software and associated documentation files (the
  * "Software"), to deal in the Software without restriction, including
@@ -11,10 +11,10 @@
  * distribute, sublicense, and/or sell copies of the Software, and to
  * permit persons to whom the Software is furnished to do so, subject to
  * the following conditions:
- * 
+ *
  * The above copyright notice and this permission notice shall be
  * included in all copies or substantial portions of the Software.
- * 
+ *
  * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
  * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
  * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
@@ -22,7 +22,7 @@
  * LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
  * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
  * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
- * 
+ *
  */
 package com.udojava.evalex;
 
@@ -44,7 +44,7 @@ import com.udojava.evalex.Expression.LazyNumber;
 
 /**
  * <h1>EvalEx - Java Expression Evaluator</h1>
- * 
+ *
  * <h2>Introduction</h2> EvalEx is a handy expression evaluator for Java, that
  * allows to evaluate simple mathematical and boolean expressions. <br>
  * For more information, see:
@@ -61,7 +61,7 @@ import com.udojava.evalex.Expression.LazyNumber;
  * "http://www.amazon.de/Java-Number-Cruncher-Programmers-Numerical/dp/0130460419">The
  * Java Programmers Guide To numerical Computing</a> (Ronald Mak, 2002).</li>
  * </ul>
- * 
+ *
  * @authors Thanks to all who contributed to this project: <a href=
  *          "https://github.com/uklimaschewski/EvalEx/graphs/contributors">Contributors</a>
  * @see <a href="https://github.com/uklimaschewski/EvalEx">GitHub repository</a>
@@ -285,7 +285,7 @@ public class Expression {
 
 		/**
 		 * Creates a new operator.
-		 * 
+		 *
 		 * @param oper
 		 *            The operator name (pattern).
 		 * @param precedence
@@ -302,7 +302,7 @@ public class Expression {
 
 		/**
 		 * Creates a new operator.
-		 * 
+		 *
 		 * @param oper
 		 *            The operator name (pattern).
 		 * @param precedence
@@ -376,7 +376,7 @@ public class Expression {
 
 		/**
 		 * Creates a new tokenizer for an expression.
-		 * 
+		 *
 		 * @param input
 		 *            The expression string.
 		 */
@@ -391,7 +391,7 @@ public class Expression {
 
 		/**
 		 * Peek at the next character, without advancing the iterator.
-		 * 
+		 *
 		 * @return The next character or character 0, if at end of string.
 		 */
 		private char peekNextChar() {
@@ -519,7 +519,7 @@ public class Expression {
 	/**
 	 * Creates a new expression instance from an expression string with a given
 	 * default match context of {@link MathContext#DECIMAL32}.
-	 * 
+	 *
 	 * @param expression
 	 *            The expression. E.g. <code>"2.4*sin(3)/(2-4)"</code> or
 	 *            <code>"sin(y)>0 & max(z, 3)>3"</code>
@@ -531,7 +531,7 @@ public class Expression {
 	/**
 	 * Creates a new expression instance from an expression string with a given
 	 * default match context.
-	 * 
+	 *
 	 * @param expression
 	 *            The expression. E.g. <code>"2.4*sin(3)/(2-4)"</code> or
 	 *            <code>"sin(y)>0 & max(z, 3)>3"</code>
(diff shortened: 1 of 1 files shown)


RobertZenz:
You've got quite a few whitespace-changes in there which do not seem to be needed, was that intentional? If not, could you clean them out?

I'm not familiar with such mathematical constructs, why the test against `1`?

ayanovsk:
I used the input values which I showed above and printed out why it does not break out of the while loop for these values, and it showed that Ix and ixprev differ by one and flip between ix greater than ixprev by one and ixprev greater than ix by one. I understand that for certain values it does not go down to exact equality, but rather ends with difference of one. Maybe I'm not right; not absolutely sure, but it really goes into infinite loop otherwise, and was working correctly for those cases, which I tried after this correction. 

I will look at white spaces tomorrow; will try to understand and address what is the issue with those. I think I added two lines and replaced one line.

RobertZenz:
I think your editor is set to automatically remove trailing whitespace, which seems to be the issue here.

ayanovsk:
Yes, it probably does. 
What is the rule to follow about trailing white spaces? Do we want to keep all lines of the same length?

RobertZenz:
My simple rule for contributing to projects is: Change nothing that you don't need to change, the diff must contain only the expected changes.

That said, I believe EvalEx does not have any style guide, so "as it currently is" applies.

RobertZenz:
You can simply force push (`git push --force YOURREPO master`) and the pull request will automatically update.

ayanovsk:
Thanks a lot, I think I understand; will try to do a new commit on my branch and then will force push it tomorrow. Will communicate it then here. 

By the way, I could not compile the whole thing as it is, as it complained about not final variables v1 and v2 in the inner class in two Java files. Strange, was I doing something wrong? I could compile in a separate location to test my change in Expression.java only after I made those fields final. Of course did not touch those things in the commit, were I only changed those three lines in Expression.java, as you saw.

RobertZenz:
With "a completely new commit" I meant that that is *replacing* the current one.

What IDE are you using? This is a compiler setting and you should be able to turn it off somewhere in your IDE preferences.

ayanovsk:
So you mean, it should be still --amend, not completely new commit? I thought, I can make a new commit in my branch, but then push --force will replace the current commit in my fork in github?

I am using IntellijIdea for actual work, but in this case I just modified these three lines in Expression.java using TextPad (such simple windows editor) and then just copied over all .java files to my IntellijIdea project to make sure it runs OK (that is where I encountered those v1 and v2 fields which did not compile unless I made them final).  But the changed Expression.java file in the commit was modified by TextPad.

RobertZenz:
> So you mean, it should be still --amend, not completely new commit?

As it it suits you.

 > I thought, I can make a new commit in my branch, but then push --force will replace the current commit in my fork in github?

*Just* creating a new commit on top of the old one won't get rid of the old one. You'd have to remove the old one somehow, for example rebasing. Or you revert your branch to the latest master, do a new commit and then push force.

 > I am using IntellijIdea for actual work, but in this case I just modified these three lines in Expression.java using TextPad (such simple windows editor) and then just copied over all .java files to my IntellijIdea project to make sure it runs OK (that is where I encountered those v1 and v2 fields which did not compile unless I made them final). But the changed Expression.java file in the commit was modified by TextPad.

I've never used IntelliJ, so you'd have to check how to change the compiler settings there. You can also use the attached maven buildfile, if you've got Maven laying around. And otherwise tell TextPad that it should not strip whitespace.

RobertZenz:
> Now I am kind of not sure I understand. Can we abandon this pull request? How can we do it?

No, you can update your branches and reuse this pull request. Sorry for the short explanations on my end.

 > I can then checkout from my master which has not been touched to another branch and create a brand new commit and push to another branch and make new pull request from that branch.

Yes, with the exception that you push to *this* branch, this way the pull request will be reused (because you've changed the `sqrt_modification` on the server). Does that make more sense to you?

ayanovsk:
why is it so necessary to use this branch - it is only on my fork, why can't I create a new fork and make new pull request? As of now this my wrong branch has not affected anything other than in my own fork. Why can't we just forget about it?
That would be much more understandable for me.

RobertZenz:
> Why is it so necessary to use this branch...

Mh, it isn't at all. I've just the habit (and I believe the "main Github workflow", too) of reusing the same pull request when possible/sensible (keeps the repository a little cleaner and you see the history of it in the comments). But it is no necessity, no. If you are more comfortable by creating a new one, go for that, I did not wish to "impose" any of that on you.

ayanovsk:
Thanks a lot, Robert
As it is simpler to understand let me start with this way. Please decline the existing pull request. I will create a new one hopefully later today. I will pay attention to white spaces, run diff and make sure that any lines I did not need to change are not affected. I am sure I will find some editor which will do it for me. 
Thanks a lot.

RobertZenz:
Before I forget about it, I believe you could shorten the condition to something like this:

    while(ix.subtract(ixPrev).abs().intValue() <= 1);

Or some such, kinda in a hurry right now, sorry.

ayanovsk:
OK, thanks a lot, I'll close this pull request when I open the new one.

Yes, that is a good advice to make the code shorter; I'll do it like that.

It would be also good if someone looks at the essence of my change.
I cannot say that I fully understand the algorithm. I am almost sure that there is an issue because 
it does not return from SQRT() for certain arguments (that can be easily checked using my examples) - and I encountered this situation several times not just for the numbers which I quote, so it has to be addressed.
When I dug into these examples I saw that ix and ixPrev are alternating in such cases with the difference of 1 and -1 and  that is what prevents from exiting from the loop, as it expects exactly zero.
It would still be good if someone with full understanding of the algorithm could look at it and confirm that this is the right way to correct the issue, or maybe suggest some other way, if this one is not right.

ayanovsk:
You know, Robert, it sounds strange, but with your piece of code the results tend to be a few digits different from both original and when I use my awkward IF statement. Do not know why, but I decided to leave my awkward IF statement. These are the results:

This is with original code   while (ix.compareTo(ixPrev) != 0 );
----
ex1 = 5.0009999
ex2 = 2.4218732
ex3 = 2.4218748
ex4 = 2.4218748
Math.sqrt for ex5: 2.421875017419355
----
then original code hanged...

This is with my IF statement   while (test.compareTo(BigInteger.ZERO) != 0 && test.compareTo(BigInteger.ONE) != 0 );
---
ex1 = 5.0009999
ex2 = 2.4218732
ex3 = 2.4218748
ex4 = 2.4218748
Math.sqrt for ex5: 2.421875017419355
ex5 = 2.421875

Process finished with exit code 0
-----

This is with while(ix.subtract(ixPrev).abs().intValue() <= 1);   :
ex1 = 5.2188334
ex2 = 2.5517447
ex3 = 2.5517459
ex4 = 2.5517459
Math.sqrt for ex5: 2.421875017419355
ex5 = 2.5517461

Process finished with exit code 0
----
Actually sqrt(5.01) in this case is substantially off.
Don't know why this becomes such a difference?
Would really appreciate if someone else looks at the essence of  it.

this was the code I was using:

  public static void main(String[] args) {

 

        Expression ex1 = new Expression("sqrt(25.01)");
        System.out.println("ex1 = " +ex1.eval());

        Expression ex2 = new Expression("sqrt(5.86547)");
        System.out.println("ex2 = " +ex2.eval());

        Expression ex3 = new Expression("SQRT(5.8654785)");
        System.out.println("ex3 = " + ex3.eval());

        Expression ex4 = new Expression("SQRT(5.8654784912345)");
        System.out.println("ex4 = " + ex4.eval());

        System.out.println("Math.sqrt for ex5: " + Math.sqrt(5.8654786));

 

        Expression ex5 = new Expression("SQRT(5.8654786)");
        System.out.println("ex5 = " + ex5.eval());

         System.exit(0);


    }

So, I'll keep my yesterday's code.


Found that VisualStudioCode editor does not add any additional white spaces. Will create new pull request soon.

ayanovsk:
replaced this one with #166 without additional white space lines which I placed in Expression.java #165 inadvertently

PR #166, opened 2018-05-16
modifies exit condition from the while loop for SQRT() calculation to…

ayanovsk:
… avoid infinite loops

Like #165 , but removed unnecessary numerous additional white space lines

### DIFF
--- a/src/main/java/com/udojava/evalex/Expression.java
+++ b/src/main/java/com/udojava/evalex/Expression.java
@@ -1045,13 +1045,14 @@ public class Expression {
 				int bits = (n.bitLength() + 1) >> 1;
 				BigInteger ix = n.shiftRight(bits);
 				BigInteger ixPrev;
-
+				BigInteger test;
 				do {
 					ixPrev = ix;
 					ix = ix.add(n.divide(ix)).shiftRight(1);
 					// Give other threads a chance to work;
 					Thread.yield();
-				} while (ix.compareTo(ixPrev) != 0);
+					test = ix.subtract(ixPrev).abs();
+				} while (test.compareTo(BigInteger.ZERO) != 0 && test.compareTo(BigInteger.ONE) != 0 );
 
 				return new BigDecimal(ix, mc.getPrecision());
 			}


ayanovsk:
Yes, this alternation 
...
24218751   24218750
24218750   24218751
...
is actually what I was seeing as well, whereas with majority of other value of input argument
they tend to converge to exact equality. 
Still when I tried to use SQRT() in more or less sophisticated calculation
which used this operation I guess hundreds of times with various arguments, then
once in a several runs with different input values, the run never finished. 
It is understandable that it may or may not happen depending on precision.
I agree, maybe it needs to be addressed in some other way. That's why I mentioned
that it needs to be checked by someone who understands the algorithm, which I myself do not understand. So, I'm sure,  they will know what is the best way to address it.

ayanovsk:
Actually this is where I see the issue:
In this statement:

ix = ix.add(n.divide(ix)).shiftRight(1);

division of two BigIntegers ends up with truncation and there is where
the problem happens:
In our case n = 586547900000000 and it is exactly equal to 24218750 x 24218752
so the algorithm divides n by ix then calculates the arithmetic average of the result and the divider and considers it to be next approximation of the square root of n.
So if ix happens to be 24218751 so when it calculates n.divide(ix) the result happens to be 
24218750.99999995870967912424550713
but because it is BigInteger division, it truncates this result to 24218750 instead of rounding it to 24218751. As a result,  if ix = 24218751 the next iteration will yields (24218751 + 24218750)/2 and this results after one more truncation in 24218750. When we feed 24218750 in the next iteration it produces quite correctly and  exactly (24218750 + 24218752)/2 = 24218751, now we feed 24218751 and get back again 24218750 because of above truncation, then we feed 24218750 and get back 24218751, etc etc
So the problem happens when n.divide(ix) produces  24218750.99999995870967912424550713 and is treated as 24218750 instead of 24218751. 
So this line:

ix = ix.add(n.divide(ix)).shiftRight(1);

should be replaced with this:

ix =  ix.add((new BigDecimal(n).divide(new BigDecimal(ix), MathContext.DECIMAL128)).setScale(0, RoundingMode.HALF_UP).toBigInteger()).shiftRight(1);

so that it should round BigInterger division instead of truncating it.
Don't know maybe it can be done in somewhat more simple way, but in any case  it should be rounded, not truncated.

PR #170, opened 2018-05-29
Added FACT(expr) function. Fixed AbstractOperator and AbstractUnaryOp…

DesigningKnights:
…erator with final for LazyNumber.

Redid the commit and got rid of all the mess in the diffs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ dependencies {
   <tr><td>ATANH(<i>expression</i>)</td><td>Returns the angle of hyperbolic tangens of a value</td></tr>
   <tr><td>RAD(<i>expression</i>)</td><td>Converts an angle measured in degrees to an approximately equivalent angle measured in radians</td></tr>
   <tr><td>DEG(<i>expression</i>)</td><td>Converts an angle measured in radians to an approximately equivalent angle measured in degrees</td></tr>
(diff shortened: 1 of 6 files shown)


RobertZenz:
Thank you very much, that is a lot easier to review.

There are still the two `final`s changes which seem superfluous.

@uklimaschewski With the exception of the `final`s being added, looks good to me. The implementation is a little bit "naive", as in there are [faster implementations available](http://www.luschny.de/math/factorial/FastFactorialFunctions.htm), but this seems okay to me, especially as the user still has the ability to override this with their own implementation, if needed (and if they already have their own, this will not impact them).

DesigningKnights:
IDEA threw an error saying that the finals are required as they are accessed from in an inner class, refused to compile without them."Variable v1 is accessed from within inner class, needs to be declared final". is the exact message.  
error final

DesigningKnights:
As for the algorithm, I went for the simplest method, avoiding recursion. I don't like recursion because it can quickly cause an out of memory error, especially if someone tries to do something like FACT(1000).

wxisme:
@DesigningKnights I think you should create a new branch to submit your PR instead of modifying the master branch directly. In addition I have submitted another PR(#176) specifically to solve the problem of `finals`.
@RobertZenz

PR #199, opened 2019-03-29
continuous-unary

tinylamb:
it's continuous-unary Operation
but not pre/post increment

### DIFF
--- a/src/main/java/com/udojava/evalex/Expression.java
+++ b/src/main/java/com/udojava/evalex/Expression.java
@@ -1906,13 +1906,8 @@ public class Expression {
 		Tokenizer tokenizer = new Tokenizer(expression);
 		while (tokenizer.hasNext()) {
 			Token token = tokenizer.next();
-			StringBuilder sb = new StringBuilder();
-			sb.append("{")
-				.append(token.type)
-				.append(":")
-				.append(token.surface)
-				.append("}");
-			infix.add(sb.toString());
+			String infixNotation = "{" + token.type + ":" + token.surface + "}";
+			infix.add(infixNotation);
 		}
 		return infix;
 	}
--- a/src/main/java/com/udojava/evalex/Expression.java
+++ b/src/main/java/com/udojava/evalex/Expression.java
@@ -498,7 +498,8 @@ public class Expression {
 				}
 
 				if (previousToken == null || previousToken.type == TokenType.OPERATOR
-						|| previousToken.type == TokenType.OPEN_PAREN || previousToken.type == TokenType.COMMA) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/udojava/evalex/Expression.java
```diff
@@ -1899,4 +1901,20 @@ public boolean isBoolean() {
...
+		Tokenizer tokenizer = new Tokenizer(expression);
+		while (tokenizer.hasNext()) {
+			Token token = tokenizer.next();
+			StringBuilder sb = new StringBuilder();
+			sb.append("{")
+				.append(token.type)
+				.append(":")
+				.append(token.surface)
+				.append("}");
+			infix.add(sb.toString());
```
RobertZenz: You don't require a `StringBuilder` for simple concatenations, as both the bytecode compiler and the JIT compiler will optimize calls like these:

~~~java
String a = b + c + d + e;
~~~

Which will already compile to:

~~~
String a = new StringBuilder().append(b).append(c).append(d).append(e);
~~~

You only *require* a `StringBuilder` when building Strings *over* a loop. For example:

~~~java
StringBuilder string = new StringBuilder();

for (Item item in lotsOfItems) {
    string.append(item.toString());
}

return string.toString();
~~~

Because if you do this with normal String operations, the memory of the whole String will be copied on every iteration.

src/test/java/com/udojava/evalex/TestContinuousUnary.java
```diff
@@ -0,0 +1,35 @@
...
+ * @author samo
+ * @date 2019/03/29
+ */
+public class TestContinuousUnary {
+
+    @Test
+    public void testContinuousUnary() {
+
+        Expression expression = new Expression("++1");
+        Assert.assertEquals(expression.eval().intValue(), 1);
```
RobertZenz: You got the expected/actual arguments swapped on all these tests.

src/test/java/com/udojava/evalex/TestContinuousUnary.java
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Created by samo on 2019/3/29.
+ *
+ * @author samo
+ * @date 2019/03/29
+ */
+public class TestContinuousUnary {
```
RobertZenz: Could you add additional tests for stacking of mutiple operators? For example, what is the expected output of `-+---+++--++-1` and `1--++++---2+-+----1`?

tinylamb: I take your advice, then push another commit.

PR #202, opened 2019-03-31
operator-assoc-fix

tinylamb:
base on this tutorial, fix some operator leftAssoc assignment
[Java Operators: Precedence and Associativity](http://www.cs-fundamentals.com/java-programming/java-operators-precedence-and-associativity.php)

### DIFF
--- a/src/main/java/com/udojava/evalex/Expression.java
+++ b/src/main/java/com/udojava/evalex/Expression.java
@@ -600,7 +600,7 @@ public class Expression {
 				return result;
 			}
 		});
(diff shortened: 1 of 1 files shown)


RobertZenz:
Can you add unit tests which secure that behavior?

uklimaschewski:
I don't see a problem with possible mixup of the associativity of the comparison operators, compared to the Java standard. Is there any example where it might play a role?
For now, I'm closing this. If there is a need to change the code, add a unit test showing a wrong evaluation, that can be fixed by this PR.

PR #203, opened 2019-03-31
circularcheck

tinylamb:
as the following unit test
```
        @Test
	public void testNestedVars() {
		String x = "1";
		String y = "2";
		String z = "2*x + 3*y";
		String a = "2*x + 4*z";

		Expression e = new Expression(a);
		e.with("x", x);
		e.with("y", y);
		e.with("z", z);

		assertEquals("34", e.eval().toString());
	}
```
if change String x = "y + z";
e.eval() will throw java.lang.StackOverflowError.

we can add evalValidation check before eval to avoid StackOverflowError

### DIFF
--- a/src/main/java/com/udojava/evalex/Expression.java
+++ b/src/main/java/com/udojava/evalex/Expression.java
@@ -1852,12 +1852,14 @@ public class Expression {
 		final Set<String> varWithValue = new HashSet<String>();
 		int lastvarWithValueCnt = varWithValue.size();
         String lastCheckVar = "";
+        final Set<String> alreadyCheckSet = new HashSet<String>();
         final Set<String> lastCheckSet = new HashSet<String>();
 		while (!checks.isEmpty()) {
 			String check = checks.peek();
 			if (mark.equals(check)) {
+                lastCheckSet.removeAll(varWithValue);
 				if (varWithValue.size() < lastvarWithValueCnt
-                    || (varWithValue.size() == lastvarWithValueCnt && lastCheckSet.contains(lastCheckVar))) {
+                    || (varWithValue.size() == lastvarWithValueCnt && alreadyCheckSet.containsAll(lastCheckSet))) {
 					throw new ExpressionException("circular reference var : " + lastCheckVar);
 				} else {
 					lastvarWithValueCnt = varWithValue.size();
@@ -1871,7 +1873,7 @@ public class Expression {
 				}
 				LazyNumber innerVariable = variables.get(check);
 				String innerExp = innerVariable.getString();
-				if (isNumber(innerExp)) {
+				if (varWithValue.contains(check) || isNumber(innerExp)) {
 					varWithValue.add(check);
 				} else {
 					Expression exp = createEmbeddedExpression(innerExp);
@@ -1881,6 +1883,7 @@ public class Expression {
 						varWithValue.add(check);
 					} else {
                         lastCheckVar = check;
+                        alreadyCheckSet.add(check);
                         lastCheckSet.addAll(tocheck);
 						checks.addAll(tocheck);
 					}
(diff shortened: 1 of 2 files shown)


src/main/java/com/udojava/evalex/Expression.java
```diff
@@ -1782,6 +1788,51 @@ public String toRPN() {
 		return result.toString();
 	}
 
+	private void evalValidation() throws Exception {
+
+		final Set<String> circularcheck = new HashSet<String>();
+		try {
+			toRPN(circularcheck);
+		} catch (Exception e) {
+			throw e;
```
RobertZenz: Why are you catching `Exception` and simply rethrow it?

src/main/java/com/udojava/evalex/Expression.java
```diff
@@ -1272,6 +1273,11 @@ private void shuntOperators(List<Token> outputQueue, Stack<Token> stack, LazyOpe
 	 * @return The result of the expression. Trailing zeros are stripped.
 	 */
 	public BigDecimal eval() {
+		try {
+			evalValidation();
+		} catch (Exception e) {
+			throw new ExpressionException(e.getMessage());
```
RobertZenz: That is evil, that will remove the original stacktrace. Please *never* do this, because it makes debugging *so much* harder.

src/main/java/com/udojava/evalex/Expression.java
```diff
@@ -1782,6 +1788,51 @@ public String toRPN() {
...
+		final Set<String> circularcheck = new HashSet<String>();
+		try {
+			toRPN(circularcheck);
+		} catch (Exception e) {
+			throw e;
+		}
+	}
+
+	/**
+	 * circularcheck, throw Exception if circular definition
```
RobertZenz: That's a not very good method documentation. It should rather describe what the function does, and also note that it does a circular reference check. Or the method has the wrong name.

src/main/java/com/udojava/evalex/Expression.java
tinylamb: ok, let me think for a moment

src/main/java/com/udojava/evalex/Expression.java
tinylamb: I take your advice and sweep out early functions, and create a new varDefValidation function.
If expression has undefined vars or circular reference vars throw ExpressionException

tinylamb: Some cases are wrong. Let me handle them

tinylamb: I have fixed the cases, you can review it.

src/main/java/com/udojava/evalex/Expression.java
```diff
@@ -1836,6 +1840,57 @@ public String getExpression() {
...
+		checks.offer(mark);
+		final Set<String> varWithValue = new HashSet<String>();
+		int lastvarWithValueCnt = varWithValue.size();
+        String lastCheck = "";
+        final Set<String> lasttodocheck = new HashSet<String>();
+		while (!checks.isEmpty()) {
+			String check = checks.peek();
+			if (mark.equals(check)) {
+				if (varWithValue.size() < lastvarWithValueCnt ) {
+					throw new ExpressionException("circular reference var");
```
RobertZenz: If available, you should always include the name of the variable which is causing the problem.

src/main/java/com/udojava/evalex/Expression.java
```diff
@@ -1836,6 +1840,57 @@ public String getExpression() {
...
+				} else if (varWithValue.size() == lastvarWithValueCnt && lasttodocheck.contains(lastCheck)){
```
RobertZenz: You could combine this into the upper if with an "or" if I'm not mistaken.

tinylamb: Ok, I have made it more clear.

RobertZenz:
Looks good to me.

uklimaschewski:
Most unit tests failing with this PR, I'm closing it

PR #211, opened 2019-07-19
Fix bug with if nested, false condition is evaluated if it is an if f…

laurentyhuel:
Fix bug with if nested, false condition is evaluated if it is an if function
Issue : https://github.com/uklimaschewski/EvalEx/issues/210

### DIFF
--- a/src/main/java/com/udojava/evalex/Expression.java
+++ b/src/main/java/com/udojava/evalex/Expression.java
@@ -213,6 +213,25 @@ public class Expression {
 		String getString();
 	}
 
+	public class LazyNumberException implements LazyNumber{
+
+		private RuntimeException throwable;
+
+		public LazyNumberException(String msg) {
+			this.throwable = new ArithmeticException(msg);
+		}
+
+		@Override
+		public BigDecimal eval() {
+			throw throwable;
+		}
+
+		@Override
+		public String getString() {
+			return null;
+		}
+	}
+
 	/**
 	 * Construct a LazyNumber from a BigDecimal
 	 */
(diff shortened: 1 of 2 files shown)


laurentyhuel:
Hello @RobertZenz @uklimaschewski ,
Any comment about this PR ?

RobertZenz:
Sorry for not responding, I'm very busy currently.

From skimming it I'm very confused as to why you've chosen to implement it that way. The main use-case is `if(1, 2, 3/0)`, right? Looking at the code I'm confused why that *is* failing at all, as the correct result is returned as `LazyNumber` and the broken one should never be evaluated in the first place. Can you shed some light here as to why that happens?

laurentyhuel:
The failed case (before PR) `if(a=0,0,if(5/a>3,2,4))`. I added unit tests for it.

For `if` function, EvalEx always evaluate condition and return `LazyNumber` for true or false result:
`lazyParams.get(0).eval()`

In case of nested `if`, EvalEx start by manage child `if`. 

In this case it manage, in first time, `if(5/a>3,2,4)` this part, so it evaluate `5/a` and return (as explained just before) Lazy true or false result.
In second time, EvalEx manage parent if: `if(a=0,0,if(5/a>3,2,4))`.
If `a=0` then a division by zero error occurred.

Why I choose this solution ?
It's the solution I found without the need of a big refactoring on the code.

The idea is: if evaluation of child `if` condition generate an exception, then return a `LazyNumberException`.
After the `if` parent is evaluated, 
- if the result is the nested `if` then LazyNumberException eval return the exception
- otherwise `LazyNumberException` eval is not called and it's other result of parent `if` which is evaluated

RobertZenz:
I finally had the time to look at this and we were both on the wrong track.

The problem is that the `IF` function directly evaluates its condition instead of returning the whole operation as a `LazyNumber`. So all you have to do is make the `IF` return a new `LazyNumber` with its operation inside instead of evaluating it directly and returning the correct lazy value. See Expression:727, there it should return a new `LazyNumber`.

Can you follow what I mean?

laurentyhuel:
@RobertZenz I created a new PR with another way to fix this bug as you suggest in your previous comments #215

Issue #210, opened 2019-07-03
Sub IF evaluate even if in evaluate false side


laurentyhuel:
Hello,

I meet an issue with EvalEx

Let us look at a concrete example:
```
IF ( a = 0, 1, IF (2/a > 1, 2, 3))
```

When a=0, the 2nd IF condition is evaluated and it crash with a division by 0, as the 2nd IF, in this case, is in the false result of the 1rst IF.

RobertZenz:
This might be related to #194.

laurentyhuel:
Yes it's seems to be same problem.

I have a look at https://github.com/gnembon/carpetmod/blob/5ec533ff6a158015bc8cdc3da028238ea6aa1ba4/src/main/java/carpet/script/Expression.java#L1353 but there's lot of changes and some breaking change with existing public method signature.

For instance, in EvalEx:
```
public Expression setVariable(String variable, BigDecimal value) {...}
public Expression setVariable(String variable, LazyNumber value) {...}
public Expression setVariable(String variable, String value) {...}
```
And in gnembon version:
```
public Expression setVariable(String variable, Value value) {}
public Expression setVariable(String variable, LazyValue value) {}
```

What do you suggest ?

RobertZenz:
If you can come up with a solution, please do so. The best course of action would be to look through all evaluations and swap *everything* that isn't already using `LazyNumber`s to that, but that's quite a piece of work. More unit tests for that would also be nice.

I'd love to do that, but I'm *very* short on time currently, that won't change for the next few months. So if you can come up with something, please do so and I'll happily review and test it.

binbinvn:
I met same problem

PR #215, opened 2019-08-12
Fix bug with if nested, false condition is evaluated if it is an if function

laurentyhuel:
Fix bug with if nested, false condition is evaluated if it is an if function , related to issue : #210

This PR  use a different way to fix this bug  that PR #211
 as asked in comment https://github.com/uklimaschewski/EvalEx/pull/211#pullrequestreview-273455041 and https://github.com/uklimaschewski/EvalEx/pull/211#issuecomment-520212442

### DIFF
--- a/src/main/java/com/udojava/evalex/LazyIfNumber.java
+++ b/src/main/java/com/udojava/evalex/LazyIfNumber.java
@@ -30,11 +30,14 @@ import com.udojava.evalex.Expression.LazyNumber;
 import java.math.BigDecimal;
 import java.util.List;
 
+/**
+ * Lazy Number for IF function created for lazily evaluated IF condition
+ */
 public class LazyIfNumber implements LazyNumber {
 
   private List<LazyNumber> lazyParams;
 
-  LazyIfNumber(List<LazyNumber> lazyParams) {
+  public LazyIfNumber(List<LazyNumber> lazyParams) {
     this.lazyParams = lazyParams;
   }
 
--- a/src/main/java/com/udojava/evalex/LazyIfNumber.java
+++ b/src/main/java/com/udojava/evalex/LazyIfNumber.java
@@ -1,0 +1,33 @@
+package com.udojava.evalex;
+
+import com.udojava.evalex.Expression.LazyNumber;
+import java.math.BigDecimal;
+import java.util.List;
+
+public class LazyIfNumber implements LazyNumber {
+
+  List<LazyNumber> lazyParams;
+
+  public LazyIfNumber(List<LazyNumber> lazyParams) {
+    this.lazyParams = lazyParams;
+  }
+
+  @Override
+  public BigDecimal eval() {
+    BigDecimal result = lazyParams.get(0).eval();
+    assertNotNull(result);
+    boolean isTrue = result.compareTo(BigDecimal.ZERO) != 0;
+    return isTrue ? lazyParams.get(1).eval() : lazyParams.get(2).eval();
+  }
+
+  @Override
+  public String getString() {
+    return "IF";
+  }
+
+  private void assertNotNull(BigDecimal v1) {
+    if (v1 == null) {
+      throw new ArithmeticException("Operand may not be null");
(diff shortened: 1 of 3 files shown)


src/main/java/com/udojava/evalex/LazyIfNumber.java
```diff
@@ -0,0 +1,33 @@
+package com.udojava.evalex;
```
RobertZenz: The license header is missing.

src/main/java/com/udojava/evalex/LazyIfNumber.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.udojava.evalex.Expression.LazyNumber;
+import java.math.BigDecimal;
+import java.util.List;
+
+public class LazyIfNumber implements LazyNumber {
+
+  List<LazyNumber> lazyParams;
```
RobertZenz: I'd make this `protected` to give subclasses a chance to access it.

src/main/java/com/udojava/evalex/LazyIfNumber.java
```diff
@@ -0,0 +1,33 @@
...
+  public BigDecimal eval() {
+    BigDecimal result = lazyParams.get(0).eval();
+    assertNotNull(result);
+    boolean isTrue = result.compareTo(BigDecimal.ZERO) != 0;
+    return isTrue ? lazyParams.get(1).eval() : lazyParams.get(2).eval();
+  }
+
+  @Override
+  public String getString() {
+    return "IF";
```
RobertZenz: I believe that other implementations are either returning an actual `String` representation of the operation, or the result as `String`. Could you double check that?

src/test/java/com/udojava/evalex/TestLazyIf.java
```diff
@@ -26,7 +26,23 @@ public void testLazyIfWithNestedFunction() {
...
+
+  @Test
+  public void testLazyIfWithNested() {
+    Expression expression = new Expression("if(a=0,0,if(5/a>3,2,4))");
+    expression.setVariable("a", new BigDecimal(0));
+    Assert.assertEquals(expression.eval(),new BigDecimal(0));
+  }
+
+  @Test(expected = ArithmeticException.class)
+  public void testLazyIfWithNested2() {
```
RobertZenz: You should name it something like `testLazyIfWithNestedFailingIf` to make it clear that this test is supposed to be the test of a failure case.

laurentyhuel: Hello @RobertZenz, thanks for your review. I made changes

src/main/java/com/udojava/evalex/LazyIfNumber.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.udojava.evalex.Expression.LazyNumber;
+import java.math.BigDecimal;
+import java.util.List;
+
+public class LazyIfNumber implements LazyNumber {
+
+  private List<LazyNumber> lazyParams;
+
+  LazyIfNumber(List<LazyNumber> lazyParams) {
```
RobertZenz: Sorry, I missed this completely on my first review. Having a public class but a package-private constructor is slightly problematic from an API point of view. As it allows everyone to see the class, but not to use it.

I'd suggest to make it `public` or the whole class `package-private`...but I'd need to do some more thinking about that one, so I'd go with `public`.

uklimaschewski:
I hope to find some time this weekend to merge this PR and have a look at the other issues.
I have just started working at a new job, so I am quite busy at the moment. 🙂

RobertZenz:
I just verified that these changes are passing all tests (including those in my project).

laurentyhuel:
@RobertZenz @uklimaschewski Could you create a new version 2.2 with this fix ?

uklimaschewski:
@laurentyhuel Version 2.2 was built and should pop up at Maven Central in the next while.

PR #227, opened 2019-10-05
Accessing operators to add an alternative operator key (to an already implemented operator)

qwert2003:
Currently the Map _operators_ is private. The following code therefore does not work (within ExtendedExpression extends Expression).
```
addOperator(new Operator("**", OPERATOR_PRECEDENCE_POWER, false) {
            @Override
            public BigDecimal eval(BigDecimal v1, BigDecimal v2) {
                return ((Operator)operators.get("^")).eval(v1, v2);
            }
        });
```

Changing private to protected solves the issue. An alternative way would be to provide a method that returns the requested Operator.

### DIFF
--- a/src/main/java/com/udojava/evalex/Expression.java
+++ b/src/main/java/com/udojava/evalex/Expression.java
@@ -159,13 +159,13 @@ public class Expression {
     /**
      * All defined functions with name and implementation.
      */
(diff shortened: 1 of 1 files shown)


uklimaschewski:
Can you also make the other maps (variables, functions) protected, too?

qwert2003:
I have made the other Maps "variables" and "functions" protected too. Also a few useful methods.

PR #240, opened 2020-06-16
Added Summation (SIGMA)

MrMaurice211:


uklimaschewski:
Sorry for the delay, but is this of any general use? Apart of that it is missing proper documentation of what it does exactly, I see no need to put this specialized function to the standard library of functions.

MrMaurice211:
> Sorry for the delay, but is this of any general use? Apart of that it is missing proper documentation of what it does exactly, I see no need to put this specialized function to the standard library of functions.

Well is just another arithmetical expression which i think can fit pretty well, this is the one:

https://en.wikipedia.org/wiki/Summation

RobertZenz:
Here's a unittest and documentation missing, so that we can see the usecase.

MrMaurice211:
Currently not working, incompatible with the current variable detection system, thanks for looking at it!

Issue #236, opened 2020-03-19
Easier Absolute Value and Factorial


NumberC:
Hello! This is a feature request. Instead of having to write `FACT(x)` or `ABS(x)`, is it possible to add a feature where users can write `x!` or `|x|` and return the same result? I'm willing to help if needed. Thank you!

RobertZenz:
This is partially a duplicate of #126, which is concerned with right-hand unary operators, and maybe #180.

But I don't believe that there is a ticket for having an operator *around* the variable.

RobertZenz:
Oh, and to rehash previous discussions about this, the parser does not support either case as far as I am aware.

uklimaschewski:
I think this will be too complex to implement, especially that |a| Syntax is far off the map.

uklimaschewski:
Fixed with #257

Issue #180, opened 2018-09-09
Percentage operator


vanniktech:
I'd like to have a percentage operator. Initially, I thought of using `%` but seems like this library is already reserving it for modulo.

Some example input and what I'd expect.

```
5% = 0.05
5%x5 = 0.25
500% = 5
500%10 = 50
51%6.75=3.4425
```

Maybe a dedicated function `perc` could help here since `%` is already taken.

danielpeintner:
out of curiosity (and also need)... @vanniktech did you look into the internals of EvalEx to check whether it might be possible to implement the percentage operator.

I am aware of modulo and the possible conflicts but to my understanding it would be ideal to allow both "2 % 3" as modulo expression and "20%" to match peoples expectation.

RobertZenz:
If I recall correctly, for the parser it does not matter whether it is unary or normal operator. So this would conflict for sure.

uklimaschewski:
Fixed with #257

Issue #126, opened 2017-09-18
Left unary operator


xLukasbx:
Hi,
We can used to unary operator to right side example 5% at the moment? I see in tests that we can used to unary operator but to left side for instance -5.

uklimaschewski:
No, this is not possible. Unary operators are only prefix operators.

xLukasbx:
I think that it is good feature that improvement this library about unary operator on right side in the future.

uklimaschewski:
I marked this as an enhancement.

xLukasbx:
Any have idea how start add feature right side unary operator? I must admit that it is a little difficult for me. I will be grateful for any instruction.

uklimaschewski:
This is probably not so easy. Without knowing, if this is the best or right way:
1) The UnaryOperator class needs a property, if it is postfix or prefix.
2) The parser needs to be adopted to understand the new syntax
3) The shunting yard has to be extended to hanlde a postix unary operator.

uklimaschewski:
Fixed with #257

PR #255, opened 2021-04-20
Postfix unary operator

luiz-resende:
Hi,

First, I would like to thank you for the work. I have recently implemented a calculator (still to be uploaded) and was looking for an expression evaluator that could surpass the limitations of double-precision with BigDecimal, and I think the work shown in EvalEx is excellent.

So, I saw that this subject has already been discussed on issues #126, #180 and #236, but since my previous code already used custom/user-defined operators, I went ahead and decided to modify the AbstractOperator and Expression classes, so they were able to handle **postfix unary operators** (e.g. "!" operator for factorial function) without throwing exceptions. However, I did not exclude the AbstractUnaryOperator class or joined it with the AbstractOperator class through the modifications herein. Therefore, the **prefix unary operators** are still defined by the AbstractUnaryOperator class.

In general lines, I added an `int` parameter in the AbstractOperator and AbstractLazyOperator class methods called "numberOperands" to keep track of the number of operands this operator expects (e.g. numberOperands=1 for "!" and numberOperands=2 for "^"). The method ```getNumberOperands()``` below was added to the LazyOperator and AbstractLazyOperator classes, associated with a ```protected int numberOperands``` variable. This value could later be called and used whenever necessary to check which type of operator is being used.

```java
public int getNumberOperands(){
	return numberOperands;
}
```

With the above, the Expression class was edited at some points to understand and handle double-operand operators and postfix unary operators based on this ```numberOperands``` parameter, still classifying all as ```TokenType.OPERATOR```, accepting the second operand ```v2=null``` without exceptions. As an example, when attributing the token types, a condition was added such that a postfix unary operator followed by a regular operator would not classify the later as a *prefix unary operator*, i.e. when analyzing the expression ```"4!+2"``` the operator "+" would not be classified as ```TokenType.UNARY_OPERATOR``` but rather ```TokenType.OPERATOR``` accepting it as a *postfix unary operator*, yielding the correct result ```4!+2 = 26```. It is crucial to notice that a sufficiently high precedence number must be passed when declaring the new AbstractOperator to avoid misinterpretations (e.g. I used precedence "Expression.OPERATOR_PRECEDENCE_POWER_HIGHER + 1" to the added postfix unary operators).

The expression evaluation has shown the correct results for the tests I have run so far, with different expressions and postfix unary operators. However, if any mistakes were made or some logical error I have not noticed, please let me know. For now, I didn't want to change the implementation for the prefix unary operators, but if the solution presented holds, I would gladly make changes to join all operators in a single class.

Again, thank you for the repository. Regards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,6 @@ e.addOperator(new AbstractOperator("!", Expression.OPERATOR_PRECEDENCE_POWER_HIG
 	public BigDecimal eval(BigDecimal v1, BigDecimal v2) {
 		Expression.assertNotNull(v1);
 		BigDecimal fact = v1.subtract(BigDecimal.ONE);
-		BigDecimal counter = 
 		if(v1.compareTo(BigDecimal.ZERO) == 0 || v1.compareTo(BigDecimal.ONE) == 0) {
 	    		return new BigDecimal(BigDecimal.ONE);
 		}
--- a/README.md
+++ b/README.md
@@ -246,22 +246,22 @@ Or another example, add an operator `n!`, that calculates the factorial of n:
 Expression e = new Expression("4!");
 
 e.addOperator(new AbstractOperator("!", Expression.OPERATOR_PRECEDENCE_POWER_HIGHER + 1, true, 1) {
-	@Override
-	public BigDecimal eval(BigDecimal v1, BigDecimal v2) {
-		if (v1 == null) {
-			throw new ArithmeticException("Operand may not be null");
-		}
-		BigDecimal fact = v1.subtract(BigDecimal.ONE);
-		if(v1.compareTo(BigDecimal.ZERO) == 0 || v1.compareTo(BigDecimal.ONE) == 0) {
-			return BigDecimal.ONE;
-		} else {
-			while(fact.compareTo(BigDecimal.ONE) > 0) {
-				v1 = v1.multiply(fact);
-				fact = fact.subtract(BigDecimal.ONE);
-			}
-			return v1;
-		}
-	}
+    @Override
+    public BigDecimal eval(BigDecimal v1, BigDecimal v2) {
+        if (v1 == null) {
+            throw new ArithmeticException("Operand may not be null");
+        }
+        BigDecimal fact = v1.subtract(BigDecimal.ONE);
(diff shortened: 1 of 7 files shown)


RobertZenz:
I've only quickly looked over the description, and it does sound good. However, `getNumberOperands` returning an `int` seems like a not so good choice, given that it can only be 1 or 2 (also I'd rename it to `getNumberOfOperands` or `getOperandsNumber`). I don't have in my head right now how EvalEx works in this regard, but maybe an Enum would be better suited, something like UNARY_LEFT/UNARY_RIGHT/BINARY or something like that.

But as I've said, I've only skipped over it and currently can't remember how the library works in ths regard.

luiz-resende:
Hi Robert. I understand the number of operands has only two choices (1 or 2), but I think returning ```int``` is an easier and clean way (besides intuitive) of understanding and using this number within the code, e.g. in Expression.java I used the ```int``` directly return from ```getNumberOperands``` in IF conditions and operations where ```int``` is required, which I found faster and more intuitive.

Moreover, I have to say that, after looking at the code, my first instinct was also to work with the Enum ```OPERATOR``` and ```UNARY_OPERATOR``` types, creating a new one (I called it ```POST_UNARY```), so the changes would be contained to the Expression.java. However, this showed to be **more complicated and prone to error** given the structure and way the ```TokenType``` is analyzed and attributed, with more IF-ELSE blocks required and misclassification of operator's token and consequent errors occurring. With the new declared argument ```numberOperands``` one has another feature that can be used to compare and correctly analyze and evaluate the desired expression without necessarily creating a *third* ```TokenType``` for operators.

Therefore, I think a new ```int``` argument with a method returning its ```int``` value is an easy, simple and fitting solution.

Lastly, I don't see the difference from naming the method ```getNumberOperands``` to ```getOperandsNumber```, or even ```getNumberOfOperands``` since they all carry the same meaning and can be all perfectly understood to what their function is. But if you think one of the later would best fit the style used in the repository, I can change it for sure :)

uklimaschewski:
Thank you for the PR, but there are build errors with this PR. Can you please also squash the 7 commits into one and resubmit?

PR #257, opened 2021-05-06
Unary postfix  operator (squashed commits and fix #255)

luiz-resende:
Hi,

I closed the pull request #255 and, as per the author's request, squashed all commits.

The build errors seemed to be caused because the files ```TestCustoms.java```, ```TestExposedComponents.java```, ```TestNested.java``` and ```TestTokenizer.java``` were not updated with the necessary information for the ```new Operator()``` constructors. This has now been fixed.

Regards.

############################## ORIGINAL MESSAGE ##############################

So, I saw that this subject has already been discussed on issues #126, #180 and #236, but since my previous code already used custom/user-defined operators, I went ahead and decided to modify the AbstractOperator and Expression classes, so they were able to handle **postfix unary operators** (e.g. "!" operator for factorial function) without throwing exceptions. However, I did not exclude the AbstractUnaryOperator class or joined it with the AbstractOperator class through the modifications herein. Therefore, the **prefix unary operators** are still defined by the AbstractUnaryOperator class.

In general lines, I added an `int` parameter in the AbstractOperator and AbstractLazyOperator class methods called "numberOperands" to keep track of the number of operands this operator expects (e.g. numberOperands=1 for "!" and numberOperands=2 for "^"). The method ```getNumberOperands()``` below was added to the LazyOperator and AbstractLazyOperator classes, associated with a ```protected int numberOperands``` variable. This value could later be called and used whenever necessary to check which type of operator is being used.

```java
public int getNumberOfOperands(){
	return numberOperands;
}
```

With the above, the Expression class was edited at some points to understand and handle double-operand operators and postfix unary operators based on this ```numberOperands``` parameter, still classifying all as ```TokenType.OPERATOR```, accepting the second operand ```v2=null``` without exceptions. As an example, when attributing the token types, a condition was added such that a postfix unary operator followed by a regular operator would not classify the later as a *prefix unary operator*, i.e. when analyzing the expression ```"4!+2"``` the operator "+" would not be classified as ```TokenType.UNARY_OPERATOR``` but rather ```TokenType.OPERATOR``` accepting it as a *postfix unary operator*, yielding the correct result ```4!+2 = 26```. It is crucial to notice that a sufficiently high precedence number must be passed when declaring the new AbstractOperator to avoid misinterpretations (e.g. I used precedence "Expression.OPERATOR_PRECEDENCE_POWER_HIGHER + 1" to the added postfix unary operators).

The expression evaluation has shown the correct results for the tests I have run so far, with different expressions and postfix unary operators. However, if any mistakes were made or some logical error I have not noticed, please let me know. For now, I didn't want to change the implementation for the prefix unary operators, but if the solution presented holds, I would gladly make changes to join all operators in a single class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,37 @@ e.addOperator(new AbstractOperator(">>", 30, true) {
 e.eval(); // returns 212.34
 ````
 
+Or another example, add a postfix unary operator `n!`, that calculates the factorial of n. The parameters for postfix unary operators are the operator's string, its precedence, if it is left associative, is it is boolean and if it is unary (<code>true</code>).
+
+````java
+Expression e = new Expression("4!");
+
+e.addOperator(new AbstractOperator("!", Expression.OPERATOR_PRECEDENCE_POWER_HIGHER + 1, true, false, true) {
+    @Override
+    public BigDecimal eval(BigDecimal v1, BigDecimal v2) {
+        if(v1 == null) {
+            throw new ArithmeticException("Operand may not be null");
+        }
+        if(v1.remainder(BigDecimal.ONE) != BigDecimal.ZERO) {
+            throw new ArithmeticException("Operand must be an integer");
+        }
+        BigDecimal factorial = v1;
+        v1 = v1.subtract(BigDecimal.ONE);
+        if (factorial.compareTo(BigDecimal.ZERO) == 0 || factorial.compareTo(BigDecimal.ONE) == 0) {
+            return BigDecimal.ONE;
+        } else {
+            while (v1.compareTo(BigDecimal.ONE) > 0) {
+                factorial = factorial.multiply(v1);
+                v1 = v1.subtract(BigDecimal.ONE);
+            }
+            return factorial;
+        }
+    }
+});
+
+e.eval(); // returns 24
+````
+
 ### Add Custom Functions
 
 Adding custom functions is as easy as adding custom operators. Create an instance of `Expression.Function`and add it to the expression.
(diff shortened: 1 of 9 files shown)


uklimaschewski:
Thank you, this looks a bit better now, though there are still 2 commits.

I have a problem with this PR, as it changes the public interface for adding operators.
Existing systems using EvalEx with custom operators will have a problem, when upgrading to a new version using the new interface.
That's also why the unit tests are there and should not be changed, unless there is a consent to break the API.

I would propose to keep the original constructors, with a default number of parameters of 2 and adding an additional constructor with the number of operands parameter.

When thinking further, the question arises if there is an operand with other than one or two operands?
So the parameter and the getter could just be a boolean ``isUnary``, which would make things clearer.
e.g.:

`if(operators.get(token.surface).getNumberOfOperands() < 2)`

would read much better with:

`if(operators.get(token.surface).isUnary())`

Also, please remove the "CODE MODIFICATION" comment. If there is a comment needed in the code, it can stay. But without that "CODE_MODIFICATION" header and story what was changed. Code change history is maintained by Git, not in the code.

luiz-resende:
Hello,

Thank you for the idea. I have to be honest and say that I tackled the modifications as a first-time user, and the support for current users slipped my mind. I think that is now corrected. I changed the default constructors to their original state and added a third constructor to handle unary custom operators. In the same way, I have reversed the modifications to the tests, only adding a new test for postfix unary custom operators. All in just 1 commit now.

I chose to go directly to the use of a Boolean parameter <code>unaryOperator</code> and a Boolean method <code>isUnaryOperator()</code> to get it, as you mentioned. My previous choice was based on the use of the <code>int</code> value in the ```Expression.java```, but this was easily fixed with a local variable.

Regarding the "CODE MODIFICATION" in the comments, I know Git maintains a change history. This was for my use only when editing the code (outside Git), to quickly find the places where changes were made and needed to be revised. I forgot to delete it when pasting the code.

Regards.

Issue #244, opened 2020-07-06
String variables in functions


eduardo-moreira:
When a function has a String variable as param, eval consider the name of variable.

For eg:

```
Expression expression = new Expression("IN (myValue, \"CC\", \"ABC\")");
expression = expression.with("myValue", "CC");
```

It will return 1, but 0 (ZERO) is expected.

To fix it, we can change the Expression class, method eval:

```
case VARIABLE:
                    if (!variables.containsKey(token.surface)) {
                        throw new ExpressionException("Unknown operator or function: " + token);
                    }

                    stack.push(new LazyNumber() {

                        public BigDecimal eval() {
                            LazyNumber lazyVariable = variables.get(token.surface);
                            BigDecimal value = lazyVariable == null ? null : lazyVariable.eval();
                            return value == null ? null : value.round(mc);
                        }

                        public String getString() {
                            // return token.surface;   RETURN THE NAME OF THE VAR?
                            LazyNumber lazyVariable = variables.get(token.surface);
                            return lazyVariable.getString();
                        }
                    });
```

SirError:
is there an easy way to override this method with this fix? 

> 
> 
> When a function has a String variable as param, eval consider the name of variable.
> 
> For eg:
> 
> ```
> Expression expression = new Expression("IN (myValue, \"CC\", \"ABC\")");
> expression = expression.with("myValue", "CC");
> ```
> 
> It will return 1, but 0 (ZERO) is expected.
> 
> To fix it, we can change the Expression class, method eval:
> 
> ```
> case VARIABLE:
>                     if (!variables.containsKey(token.surface)) {
>                         throw new ExpressionException("Unknown operator or function: " + token);
>                     }
> 
>                     stack.push(new LazyNumber() {
> 
>                         public BigDecimal eval() {
>                             LazyNumber lazyVariable = variables.get(token.surface);
>                             BigDecimal value = lazyVariable == null ? null : lazyVariable.eval();
>                             return value == null ? null : value.round(mc);
>                         }
> 
>                         public String getString() {
>                             // return token.surface;   RETURN THE NAME OF THE VAR?
>                             LazyNumber lazyVariable = variables.get(token.surface);
>                             return lazyVariable.getString();
>                         }
>                     });
> ```

eduardo-moreira:
I cloned the source project in my workspace and use my version of Expression class

SirError:
> 
> 
> I cloned the source project into my workspace and use my version of Expression class

[...]

PR #256, opened 2021-04-27
Fix for #244

SirError:
The fix follows what was suggested by the issue creator

PR #258, opened 2021-05-25
String variable was not recognized when passed as a param

uklimaschewski:


uklimaschewski:
Originated from PR #256, fixes issue #244

PR #270, opened 2021-10-25
Add JPMS support

MDZhB:
Alter `pom.xml` to generate a `module-info.class`. All other class files still target Java 6. This allows the use of EvalEx with jlink out of the box.

If I'm reading the documentation right, JDK 9 availability varies depending on the Travis CI environment. This means it may be necessary to change `.travis.yml` to install the right JDK version -- let me know.

PR #271, opened 2021-11-13
Fix isNumber check for empty strings

ssheikin:


uklimaschewski:
merged this, thanks!

ssheikin:
Thank you for merging!
Could I ask for a new release with this fix? 
It will allow to compare empty strings with `STREQ` from the example.

uklimaschewski:
When #270 is working, I will make a release

RobertZenz:
As a sidenote, swapping the two checks would have had the same effect:

```java
if (st.length() == 1 && st.charAt(0) == MINUS_SIGN) {
```

uklimaschewski:
@ssheikin I just created version 2.7. Might take a while, until it pops up in Maven Central.

PR #273, opened 2021-11-16
Omit `NULL` from `getUsedVariables()` output

### LINKED ISSUE #238: "NULL" should not be returned as used variable
Whenn i use "NULL" in a expression, and i call the method getUsedVariables(), it also returns "NULL" as variable.

But as the documentation says, "NULL" is the constant for null.

MDZhB:
Fixes #238. Moves all constants to a map to avoid the code duplication that causes errors of this nature. Includes tests.

### DIFF
--- a/src/main/java/com/udojava/evalex/Expression.java
+++ b/src/main/java/com/udojava/evalex/Expression.java
@@ -34,6 +34,7 @@ import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Deque;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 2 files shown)


src/main/java/com/udojava/evalex/Expression.java
```diff
@@ -30,17 +30,7 @@
...
-import java.util.Collections;
-import java.util.Deque;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
-import java.util.Stack;
-import java.util.TreeMap;
+import java.util.*;
```
uklimaschewski, line 33: Please do not use wildcard imports

uklimaschewski (changes requested):

src/main/java/com/udojava/evalex/Expression.java
MDZhB: Fixed, thanks.

uklimaschewski (approved):

PR #274, opened 2021-11-16
Fix string tokenization

### LINKED ISSUE #267: getUsedVariables() throws NPE
Note: Version of EvalEx: 2.6
Custom Functions: I have also added custom functions to handle strings following this: https://github.com/uklimaschewski/EvalEx#custom-functions-with-string-parameters

Test case: Testing simple calculation and assignment expressions and for checking what values to provide from database for used variables, I make use of the getUsedVariables() method.

1.  When there is a simple assignment expression like var1=10, the correct result is computed.  In this case getUsedVariables() returns an empty list.

2.  When there is a simple assignment expression of a string like var1="somestring", then the getUsedVariables() throws a null pointer exception.  **The data here has starting double quotes and ending double quotes.  If the quotes are not there in the data then getUsedVariables() works properly.**

Stack trace:  (only 2 lines from EvalEx are printed)

```
16:32:50,154 ERROR java.lang.NullPointerException  
16:32:50,154 ERROR at com.udojava.evalex.Expression$Tokenizer.next(Expression.java:507)  
16:32:50,155 ERROR at com.udojava.evalex.Expression.getUsedVariables(Expression.java:2078)  
16:32:50,155 ERROR com.unvired.operation.CalculateRequest.calculate(CalculateRequest.java:99)  
```

If the routine is allowed to run (removing the GetUsedVariables()) call then the following NPE occurs:

```
17:04:08,413 ERROR java.lang.NullPointerException
17:04:08,484 ERROR at com.udojava.evalex.Expression$Tokenizer.next(Expression.java:507)
17:04:08,598 ERROR at com.udojava.evalex.Expression.shuntingYard(Expression.java:1342)
17:04:08,686 ERROR at com.udojava.evalex.Expression.getRPN(Expression.java:1914)
17:04:08,761 ERROR at com.udojava.evalex.Expression.eval(Expression.java:1507)
17:04:08,859 ERROR at com.unvired.operation.CalculateRequest.calculate(CalculateRequest.java:154)
```
The expectation was that the string literal would be returned.  Is this incorrect or an invalid expectation?

MDZhB:
Fixes #267. The tokenizer previously did not consume closing quotes in string literals until the following call to `next()`. This meant that the tokenizer always behaved as though there was more input to consume after a string literal. If the last token in an expression was a string literal, upon reaching the end of the expression `hasNext()` erroneously returned `true`, and the subsequent call to `next()` produced an empty token with type null.

This PR includes some minor refactoring of tokenizer code to make it less repetitious and easier to read. Lines that manipulate the tokenizer position and token content have been replaced with the methods `consumeChar()` and `skipChar()`. It also adds a new `TokenizerException`; this inherits from `ExpressionException` so as to not break existing exception handlers.

Note that the new string logic will change the behaviour of strings slightly: literals without closing quotation marks will cause a `TokenizerException`.

### DIFF
--- a/src/main/java/com/udojava/evalex/TokenizerException.java
+++ b/src/main/java/com/udojava/evalex/TokenizerException.java
@@ -1,0 +1,7 @@
+package com.udojava.evalex;
+
+public class TokenizerException extends Expression.ExpressionException {
+    public TokenizerException(String message, int characterPosition) {
+        super(message, characterPosition);
+    }
+}
--- a/src/test/java/com/udojava/evalex/TestGetUsedVariables.java
+++ b/src/test/java/com/udojava/evalex/TestGetUsedVariables.java
@@ -3,6 +3,8 @@ package com.udojava.evalex;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
+import java.math.BigDecimal;
+import java.util.Collections;
 import java.util.List;
 import org.junit.Test;
 
@@ -32,5 +34,12 @@ public class TestGetUsedVariables {
     Expression ex = new Expression("1/2");
     List<String> usedVars = ex.getUsedVariables();
     assertEquals(0, usedVars.size());
+  }
+
+  @Test
+  public void testStringComparison() {
+    // test for issue #267 (getUsedVariables() throws NPE in expressions containing string literals)
+    Expression ex = new Expression("foo=\"bar\"").setVariable("foo", (BigDecimal) null);
+    assertEquals(Collections.singletonList("foo"), ex.getUsedVariables());
   }
 }
--- a/src/test/java/com/udojava/evalex/TestTokenizer.java
+++ b/src/test/java/com/udojava/evalex/TestTokenizer.java
@@ -1,15 +1,18 @@
 package com.udojava.evalex;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-
 import com.udojava.evalex.Expression.Token;
 import com.udojava.evalex.Expression.TokenType;
 import java.math.BigDecimal;
 import java.util.Iterator;
+
 import org.junit.Test;
+import org.junit.function.ThrowingRunnable;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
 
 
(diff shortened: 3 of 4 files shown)


src/test/java/com/udojava/evalex/TestTokenizer.java
```diff
@@ -1,15 +1,16 @@
...
 import com.udojava.evalex.Expression.TokenType;
 import java.math.BigDecimal;
 import java.util.Iterator;
+
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.function.ThrowingRunnable;
+import org.junit.rules.ExpectedException;
+
+import static org.junit.Assert.*;
```
uklimaschewski, line 13: please do not use wildcard imports

src/test/java/com/udojava/evalex/TestGetUsedVariables.java
```diff
@@ -33,4 +35,11 @@ public void testVarsNothing() {
     List<String> usedVars = ex.getUsedVariables();
     assertEquals(0, usedVars.size());
   }
+
+  @Test
+  public void testStringComparison() {
+    // test for issue #267 (getUsedVariables() throws NPE in expressions containing string literals)
+    Expression ex = new Expression("foo=\"bar\"").setVariable("foo", (BigDecimal) null);
+    assertEquals(Arrays.asList("foo"), ex.getUsedVariables());
```
uklimaschewski, line 43: replace with `Collections.singletonList("foo")`

src/test/java/com/udojava/evalex/TestTokenizer.java
```diff
@@ -1,15 +1,16 @@
...
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-
 import com.udojava.evalex.Expression.Token;
 import com.udojava.evalex.Expression.TokenType;
 import java.math.BigDecimal;
 import java.util.Iterator;
+
+import org.junit.Rule;
```
uklimaschewski, line 8: Remove unused imports (org.junit.Rule and org.junit.rules.ExpectedException)

src/main/java/com/udojava/evalex/TokenizerException.java
```diff
@@ -0,0 +1,11 @@
+package com.udojava.evalex;
+
+public class TokenizerException extends Expression.ExpressionException {
+    public TokenizerException(String message) {
```
uklimaschewski, line 4: Remove this unused constructor

uklimaschewski (changes requested):

MDZhB:
@uklimaschewski addressed your feedback, thank you.

uklimaschewski:
Thank you for that useful PR! :-)

PR #292, opened 2022-08-04
Add more constants and functions.

manhphi65:
Constants : PHI (the golden ratio).
Functions : CBRT (cube root).


RobertZenz:
Looks good to me, but I'm unsure about lifting the big-math functions. I'd rather see an 
(optional) dependency to big-math so that all functions are being used and replaced with the ones from big-math.

@uklimaschewski I'd opt for splitting this MR into one only containing PHI and one with CBRT.

uklimaschewski:
@RobertZenz I see it the same, I'd like to have an additional package (repository) evalex-bigmath in 3.0, which has a dependency to bigmath and overrides actual functions and adds some more.

uklimaschewski:
Closing this PR due to inactivity.

PR #367, opened 2023-03-21
Added handling of Null values

ida-greuelr:
Hi,

I'm a happy EvalEx user but I would like to request handling of Null values. I'm not sure if this is the right way for you, but I will tried the easiest one by forking your project.
Together with a colleague of mine I created and tested an example implementation which is represented by this PR.

Please feel free to comment or request changes.


ida-greuelr:
The CodeSmells found by SonarCloud are difficult to rate. We would explicitly need `null` return values so we could distinguish between those and "correct" return values

uklimaschewski:
Interesting idea, thank you for the pull request.
But for me it is not 100% clear, what you want to achieve.
Is it that you want to compare to a new constant `NULL`, like in `if(name == null, "unknown", name)` or `a != null`?
In that case, I think there is an easier way to do it.
I think it would be enough to add the constant, the additional data type and probably that's all.
No need to change the `getXyzValue()` to return `null`, I was able to run these tests then:
```java
  @Test
  void testNullEquals() throws ParseException, EvaluationException {
    Expression expression = new Expression("a == null");
    assertThat(expression.with("a", null).evaluate().getStringValue()).isEqualTo("true");
  }

  @Test
  void testNullNotEquals() throws ParseException, EvaluationException {
    Expression expression = new Expression("a == null");
    assertThat(expression.with("a", 99).evaluate().getStringValue()).isEqualTo("false");
  }
```

Of course, this has to be tested more in-depth with all the other datatypes and boolean functions/operators.

Or are you thinking of some other use cases?

In any case, for for a successful pull request, it would be nice to add
* A new `ExpressionEvaluatorNullTest` with some use cases for how to use null in an expression.
* Add the necessary documentation in the docs/ folder (new data type and a new chapter _Null value handling_ in _Parsing and Evaluation_. Best just before the _Exception Handling_ chapter.

ida-whitek:
I tried adding a few tests with examples on how this new null values would be used.

The general idea is indeed allowing the passing of null values and handling inside the expression by comparing to the null constant inside an if function (or maybe some sort of new coalesce function).

However, if the null is not handled properly in the expression our expectation would be that operations and functions that do not support null fail. Either with a null pointer exception or with a stricter supporting type check in the functions themselves.

uklimaschewski:
Ok, this looks better now.

I have created https://github.com/ezylang/EvalEx/pull/380 and added 2 commits on top of your pull request.
These commits fix the Sonar code smells and I have added an additional test.

To get this merged, you have to provide the documentation (in /docs) about the null value handling.

Also, I think this is a new feature, not a bug fix. So it will go to the 3.1.0 version, not to the 3.0.x version.
I have created a 3.1.0 branch, can you move your pull request to that branch instead o main?
Or simply make a new PR.

Thanks for contributing. :-)

ida-greuelr:
Hello Udo,
I will try to do this tonight. I'm currently a little bit low in time. Thanks for notifying me.

ida-greuelr:
Was a long day,
I hope it's better now. All kudos belong to @ida-whitek  for her work on this topic. I did only the plumbing.

uklimaschewski (approved):
There are some problems now with the lately merged pull request #381.
I will fix them in a separate commit, because I need to also to fix the build system.

Issue #374, opened 2023-04-01
OutOfMemoryError/memory leak on `e.<number>` input


Maxr1998:
Running the following code causes my Android app the memory usage to rise rapidly until it is interrupted or crashes.

```kotlin
val ex = Expression("e.1")
ex.evaluate()
```

It's reproducible on both 3.0.4 and 3.0.3, I haven't checked any other versions. In the debugger I noticed that the `Tokenizer.tokens` list is seemingly the culprit, containing thousands of `NUMBER_LITERAL` entries with value set to an empty string.

uklimaschewski:
Hi, I would expect this to produce a parsing error before trying to evaluate.

PR #376, opened 2023-04-06
handled nested evaluation value

PrajwalBanakar26:


uklimaschewski:
Good idea, thank you. But there are two problems wit this PR:
- Missing unit test for the new conversion.
- Code formatting is wrong. You can auto-correct the formatting by running `mvn spotless:apply`.

PrajwalBanakar26:
Created unit test case for the new conversion.
Ran `mvn spotless:apply` in order to implement the correct code format

uklimaschewski:
Hi, thank you for the PR, very much appreciated. I will merge it into the upcoming 3.1.0 release. This might take a while, because I am very busy at the moment. Before 3.1.0. Plan is to release 3.0.5 first (Bugfix for #374), then do some more work on 3.1.0 and then release it.

PR #381, opened 2023-05-06
add DATE_TIME and DURATION data types

vincentlejeune:
add date_time and duration data types as discuss here #379 
_DATE_TIME_ uses _java.time.Instant_ and _DURATION_ uses _java.time.Duration_
adding or substracting number to _DATE_TIME_ convert the number in _DURATION_ in days


src/main/java/com/ezylang/evalex/data/EvaluationValue.java
```diff
@@ -310,6 +348,44 @@ public Boolean getBooleanValue() {
     }
   }
 
+  /**
+   * Gets a {@link Instant} representation of the value. If possible and needed, a conversion will
+   * be made.
+   *
+   * <ul>
+   *   <li>Any non-zero number value will return the instant from the epoc value.
+   *   <li>Any string with the string representation of an instant (ex: <code>
```
uklimaschewski, line 357: I don't see this implemented?

docs/concepts/datatypes.md
```diff
@@ -9,14 +9,16 @@ nav_order: 2
...
-| ARRAY           | java.util.List                    |
-| STRUCTURE       | java.util.Map                     |
-| EXPRESSION_NODE | com.ezylang.evalex.parser.ASTNode |
+| Data Type       | Internal Representation            |
+|-----------------|------------------------------------|
+| NUMBER          | java.math.BigDecimal               |
+| BOOLEAN         | java.lang.Boolean                  |
+| STRING          | java.lang.String                   |
+| DATE_TIME       | java.time.Instant                  |
+ | DURATION       | java.time.Duration                 |
```
uklimaschewski, line 18: indentation

src/main/java/com/ezylang/evalex/data/EvaluationValue.java
```diff
@@ -310,6 +348,44 @@ public Boolean getBooleanValue() {
     }
   }
 
+  /**
+   * Gets a {@link Instant} representation of the value. If possible and needed, a conversion will
+   * be made.
+   *
+   * <ul>
+   *   <li>Any non-zero number value will return the instant from the epoc value.
```
uklimaschewski, line 356: I think this should return something like Instant.ofEpochMilli(((BigDecimal) value).longValue())

src/main/java/com/ezylang/evalex/data/EvaluationValue.java
```diff
@@ -310,6 +348,44 @@ public Boolean getBooleanValue() {
...
+   * <ul>
+   *   <li>Any non-zero number value will return the duration from the millisecond.
+   *   <li>Any string with the string representation of an {@link Duration} (ex: <code>
+   *       "PnDTnHnMn.nS"</code>) (case ignored) will return the current instant.
+   * </ul>
+   *
+   * @return The {@link Duration} representation of the value.
+   */
+  public Duration getDurationValue() {
+    if (Objects.requireNonNull(getDataType()) == DataType.DURATION) {
```
uklimaschewski, line 383: like in getDateTimeValue(), JavaDoc does not seem to match implementation

src/main/java/com/ezylang/evalex/functions/datetime/DateTimeFormatFunction.java
```diff
@@ -0,0 +1,45 @@
...
+
+@FunctionParameter(name = "value", isVarArg = true)
+public class DateTimeFormatFunction extends AbstractFunction {
+  @Override
+  public EvaluationValue evaluate(
+      Expression expression, Token functionToken, EvaluationValue... parameterValues) {
+    String formatted;
+    if (parameterValues.length < 2) {
+      formatted =
+          LocalDateTime.ofInstant(parameterValues[0].getDateTimeValue(), ZoneOffset.UTC).toString();
```
uklimaschewski, line 35: I'm a bit unsure about the fixed UTC offsets here and in other places, woudn't the current time zone be a better choice?

src/main/java/com/ezylang/evalex/operators/arithmetic/InfixPlusOperator.java
```diff
@@ -40,6 +41,17 @@ public EvaluationValue evaluate(
...
+      return new EvaluationValue(
+          leftOperand.getDateTimeValue().plus(rightOperand.getDurationValue()));
+    } else if (leftOperand.isDurationValue() && rightOperand.isDurationValue()) {
+      return new EvaluationValue(
+          leftOperand.getDurationValue().plus(rightOperand.getDurationValue()));
+    } else if (leftOperand.isDateTimeValue() && rightOperand.isNumberValue()) {
+      return new EvaluationValue(
+          leftOperand
+              .getDateTimeValue()
+              .plus(Duration.ofDays(rightOperand.getNumberValue().longValue())));
```
uklimaschewski, line 54: Unsure  about if days is here the right choice, some people might expect it to be milliseconds

.gitignore
```diff
@@ -29,3 +29,4 @@ target/
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
+/.sdkmanrc
```
uklimaschewski, line 32: what's that?

uklimaschewski:
Thank  you very much, @vincentlejeune!
That is a great start for supporting date/time and duration values. Very similar to what I was thinking about to implement.
I added some remarks, I am a bit unsure about the default UTC time zone at some places.
Wouldn't be the current time zone be a better choice?

With EvalEx, I am following semantic versioning (https://semver.org/). 
So, this great new feature will require a new minor version. (I don't see it breaking backward compatibility).
I have already created a branch for version 3.1.0, can you please move your pull request to this branch?

I would then merge it to the 3.1.0 branch, so more date/time or other new functionality can be added.
Once I think that 3.1.0 is ready to release, it will become the main branch and 3.0.x will go into maintenance.

src/main/java/com/ezylang/evalex/functions/datetime/DateTimeFormatFunction.java
vincentlejeune: I completely agree with you. I'm also not convinced that using fixed UTC offsets is the best approach. In fact, I've been thinking that it would be more flexible if we could have a configuration option to force a specific time zone, with the default being the system time zone. This way, users could set their preferred time zone and not have to worry about inconsistencies or inaccuracies caused by using fixed offsets. What do you think?

vincentlejeune:
Thank you for your feedback, @uklimaschewski . I'm glad you found my pull request helpful. I agree with your suggestion that we should consider using the current time zone instead of UTC. In fact, I think it would be great if we could have a configuration option that allows users to force a specific time zone, with the default being the system time zone.

Regarding your request to review the pull request on the 3.1 branch, I'm happy to do that. I'll make sure to update the pull request accordingly.

One thing I wanted to bring up is that I'm not entirely sure if storing the data as an Instant is the best approach. I wonder if using a LocalDateTime type might be more appropriate, especially if we want to support more fine-grained operations on date/time values. What do you think?

src/main/java/com/ezylang/evalex/operators/arithmetic/InfixPlusOperator.java
vincentlejeune: I think it depends on the use case. I agree that without a precise use case we can add milliseconds instead of days by default

src/main/java/com/ezylang/evalex/functions/datetime/DateTimeFormatFunction.java
uklimaschewski: Having a configuration property for the default time zone sounds good. If not set explicitly, the system default time zone could be used.

vincentlejeune:
Thank you for your feedback, @uklimaschewski . I understand your point about using Instant instead of LocalDateTime as the internal representation. Based on your explanation, I agree that using Instant provides a well-defined UTC time zone and simplifies calculations and transformations.

However, I'm still trying to figure out how to handle the current time zone parameter in the functions. The expression needs to register the functions at creation time, but the API doesn't provide a mechanism to retrieve the configuration.

Do you have any suggestions on how we can resolve this issue? I'm thinking that we could perhaps pass the current time zone as a parameter to the functions, but this might make the expressions more cumbersome to write. Alternatively, we could consider adding a way to dynamically retrieve the current time zone from the system at runtime. What do you think?

uklimaschewski:
I think the time zone is only needed, when a date is passed as a variable, or created using a function.
Return value is always an Instant.
In both cases, the time zone from the configuration can be used, if not explicitly specified.
For the creation methods(s), it could be a separate method, with an additional first parameter for the time zone.
Or, just brainstorming, we introduce another type TIMEZONE that can be passed as last optional parameter, plus a creation method for it. (Probably a bit too much).

vincentlejeune:
I've just pushed the mentioned changes. However, I'm encountering an issue with the EvaluationValue API. The constructor only accepts an object as a parameter, but I need to access the default ZoneId stored in the ExpressionConfiguration object. It seems more appropriate to use a factory that has knowledge of the configuration to create the EvaluationValue object. Without this modification, the transformation of a LocalDateTime from a structure won't be able to use the configured time zone. What do you think @uklimaschewski ?

uklimaschewski (approved):
Looks good, I think this is a good start from where further enhancements to the date time handling can be made.
I see the problem with EvaluationValue and configuration of the default time zone.
This can be solved in a separate commit, I have to think about a good backward compatible solution.
I also think about adding some more functions and enhancing the documentation.
Maybe it will be possible to release a first 3.1.0 version in a few weeks, as soon as things look stable.
Also the open pull requests for null handling and subexpressions should go into that release.

PR #394, opened 2023-07-30
adds lombok.Generated to proguard ignore warning configuration

uklimaschewski:


vanniktech (commented):
I think that the file needs a proper file ending, hence: evalex.pro

vanniktech (approved):
Perfect, thank you very much 👌

vanniktech:
Hmm has anything changed regarding the process of how the JAR is published? This is broken with 3.2.0 & 3.3.0 versions. The file can no longer be picked up.

vanniktech:
Never mind, this is an issue on Androids side.

Issue #427, opened 2024-01-12
With introduction of null, it's better to allow short circuiting (lazy evaluation) for those boolean operators


stevenylai:
I see the library now support null but perhaps it's better to also allow short circuiting for those boolean operator? Consider an example:
```
IF(v != NULL && v > 0, LOG(v), 0)
```
If `v` is set to `null`, we will get an exception saying `null` cannot be compared.

It seems that operators don't have the option to be lazy? So perhaps the workaround here is to introduce an `AND` function:
```
IF(AND(v != NULL, v > 0), LOG(v), 0)
```

uklimaschewski:
I think this can be easily implemented in the expression evaluator [here](https://github.com/ezylang/EvalEx/blob/b8b681c394e5dcae8bc3ed2a9995db7d69a3e402/src/main/java/com/ezylang/evalex/Expression.java#L124)
```java
      case INFIX_OPERATOR:
        result =
            token
                .getOperatorDefinition()
                .evaluate(
                    this,
                    token,
                    evaluateSubtree(startNode.getParameters().get(0)),
                    evaluateSubtree(startNode.getParameters().get(1)));
        break;
```
The operator type has to be checked. If it is a logical _AND_, then evaluate the first operand only.
Only, if the result is _TRUE_, evaluate also the second operand.

uklimaschewski:
Similar logic can be implemented for the logical _OR_ operator. If the first operand is _TRUE_, skip evaluating the second operand.

stevenylai:
OK. I see. I think given that user can define custom operators, we can define an additional `isLazy` parameter in `InfixOperator` interface (assuming post/prefix operators do not need the same) and implement the lazy evaluation logic only for AND/OR operators. Documentation will also have to be updated.

I can also help to make the change if this sounds reasonable.

uklimaschewski:
This is exactly what I had in mind. If there is no breaking changes, it could go to a 3.2.0 version.
Feel free to volunteer, I am always happy with people contributing.  🙂

PR #432, opened 2024-01-16
#427 Lazy operands for short-circuit evaluation of logic operators

stevenylai:
This implementation changes existing `&&` and `||` operators and it seems to be fine. Alternatively we can keep the existing operators unchanged but then we need to add tests for using custom operators. I can see currently all features such as `isLazy` from functionParameter are used at least once by one of the built-in operator/function so I last time I checked, I didn't find tests for implementing some features purely for the purpose of being used by custom operator/function.

This fixes: https://github.com/ezylang/EvalEx/issues/427


uklimaschewski (approved):

uklimaschewski:
Thank you very much!

Issue #357, opened 2023-02-25
Expression should provide a thread-safe api


xuan-nguyen-swe:
I have an expression and I need to calculate results for a lot of cases of variables.

For example,
```kotlin
val expression = Expression("a + b")

// thread 1
thread {
  expression.with("a", 1)
  expression.with("b", 1)
}

// thread 2
thread {
  expression.with("a", 2)
  expression.with("b", 2)
}

// ...

// thread n
thread {
  // ...
}
```

If I understand the implementation correctly, it will not work because `Expression` object is not safe to be used in multi threads.
Could we make it thread-safe? Maybe `Expression` should allow to build immutable snapshots with a list of variables. The snapshots should reuse the expensive computation from Expression to evaluate the result.

uklimaschewski:
In fact, an expression is thread-safe, as soon as the expression was parsed.
But the standard `MapBasedDataAccessor` which holds the variable values is not thread-safe.

There are some possibilities to achieve thread safety here. I did a small test with a custom data accessor that is using `ThreadLocal` to store the data along with the thread:
```java
package com.ezylang.evalex;

import com.ezylang.evalex.config.ExpressionConfiguration;
import com.ezylang.evalex.data.DataAccessorIfc;
import com.ezylang.evalex.data.EvaluationValue;
import com.ezylang.evalex.data.MapBasedDataAccessor;
import com.ezylang.evalex.parser.ParseException;
import org.junit.jupiter.api.Test;

import java.math.BigDecimal;
import java.security.SecureRandom;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.TimeUnit;

import static org.assertj.core.api.Assertions.assertThat;

public class ThreadLocalTest {

  class ThreadLocalDataAccess implements DataAccessorIfc {
    ThreadLocal<MapBasedDataAccessor> threadLocal =
        ThreadLocal.withInitial(MapBasedDataAccessor::new);

    @Override
    public EvaluationValue getData(String variable) {
      return threadLocal.get().getData(variable);
    }

    @Override
    public void setData(String variable, EvaluationValue value) {
      threadLocal.get().setData(variable, value);
    }
  }

  @Test
  void testThreadLocal() throws ParseException, InterruptedException {

    ExpressionConfiguration configuration =
        ExpressionConfiguration.builder().dataAccessorSupplier(ThreadLocalDataAccess::new).build();

    Expression expression = new Expression("a+b", configuration);
[...]

PR #433, opened 2024-01-23
#357 Provide a copy constructor / function

stevenylai:


uklimaschewski (approved):

uklimaschewski:
Thanks a lot. I have tested it with a multithreading test. Works now in multithreading, when you use a copy in each thread.
I'll add the test later.

Issue #435, opened 2024-02-15
Broken boolean expression evaluation in version  3.1.0


Adler95:
With version **3.1.0**, the boolean expression evaluation was broken. It seems that the problem takes effect only when scale of the BigDecimal is greater than zero.

Example (correct result) with scale 0:

```
        Expression expression = new Expression("a && b");

        EvaluationValue result = expression
                .with("a", BigDecimal.ZERO)
                .and("b", BigDecimal.ZERO)
                .evaluate();

        System.out.println(result.getBooleanValue()); // prints false
```

Example (wrong result) with scale 1:

```
        Expression expression = new Expression("a && b");
        
        EvaluationValue result = expression
                .with("a", new BigDecimal("0.0"))
                .and("b", new BigDecimal("0.0"))
                .evaluate();

        System.out.println(result.getBooleanValue()); // prints true
```

Issue #417, opened 2023-12-27
Equality depends on input type in 3.1.0


agronick:
I upgraded to EvalEx:3.1.0. In 3.0.5 this works correctly. I have an expression that looks like

```
IF(a == 70, "P", IF(a == 60, "N", a))
```
This is a user inputted string to convert 70 to P for park and 60 to N for neutral. Otherwise show the gear as a number.

On 3.1.0 I get the following results:

```
expression.with("a", 70).evaluate().value
```
Returns `P`
```
expression.with("a", 70.0).evaluate().value
```
Return `7E+1`

On 3.0.5 both `70` (int) and `70.0` (double) return `P` which is correct. This appears to be a regression.

PR #437, opened 2024-02-22
#435 Fix zero with scale to boolean (main)

stevenylai:
Fixes https://github.com/ezylang/EvalEx/issues/435 in the main branch


Adler95:
@stevenylai I don't think this is the appropriate fix, as it doesn't really solve the: https://github.com/ezylang/EvalEx/issues/435.
The problem must be somewhere in the evaluation process since EvaluationValue already contains incorrect result.

stevenylai:
It's a follow-up for https://github.com/ezylang/EvalEx/issues/417 where the fix is to replace `.equals()` with `.compareTo()`. We missed the int -> boolean conversion part from the previous fix.

Adler95:
All good. I agree with you that there is a necessity to replace .equals() with .compareTo(). I just wanted to point out that it doesn't fix the issues referenced in your PR comment. https://github.com/ezylang/EvalEx/issues/435 :)

Adler95:
@stevenylai I just did a bit more detailed review of the process (at first I thought that this method is not invoked during evaluation). I can see now that this is the proper fix. I apologize for the false information and thank you for providing the fix quickly. :)

src/main/java/com/ezylang/evalex/data/EvaluationValue.java
```diff
@@ -374,7 +374,7 @@ public String getStringValue() {
   public Boolean getBooleanValue() {
     switch (getDataType()) {
       case NUMBER:
-        return !value.equals(BigDecimal.ZERO);
+        return ((BigDecimal) value).compareTo(BigDecimal.ZERO) != 0;
```
uklimaschewski, line 377: I'd do it a little bit cleaner:
`return getNumberValue().compareTo(BigDecimal.ZERO) != 0;`

`getNumberValue()`already does the cast.

uklimaschewski (changes requested):
Thanks, you were a little bit faster than me.
I have just one request: Can you not do the cast directly, but use the `getNumberValue()` method, instead?

uklimaschewski (approved):

PR #450, opened 2024-03-23
Add functions STR_STARTS_WITH and STR_ENDS_WITH

oswaldobapvicjr:
This PR adds two string functions:
- `STR_STARTS_WITH` (returns true if the string starts with the substring)
- `STR_ENDS_WITH` (returns true if the string ends with the substring)


uklimaschewski:
There are some code formatting problems, running a `mvn spotless:apply` should fix these.

PR #461, opened 2024-04-05
New STR_FORMAT function

oswaldobapvicjr:
This PR creates a new String function `STR_FORMAT(format [, argument, ...])` which can be used to produce formatted strings. For example:

````
STR_FORMAT("Greetings from %s!", "Brazil")
Result: "Greetings from Brazil!"

"STR_FORMAT("%.2f %s", 11.1, "EUR")
Result: "11.10 EUR"
````

It can be useful to apply some formatting from numbers to strings.

### DIFF
--- a/docs/references/functions.md
+++ b/docs/references/functions.md
@@ -31,13 +31,14 @@ Available through the _ExpressionConfiguration.StandardFunctionsDictionary_ cons
 
 ### String Functions
 
-| Name                               | Description                                                           |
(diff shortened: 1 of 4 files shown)


uklimaschewski:
I merged it, though I see a problem with the (non-) usage of a locale here (and also in the already existing date functions).
To fix this problems, I will introduce a configurable locale to the expression configuration.
This will also allow to get rid of the ugly static locale setting in the test.

PR #463, opened 2024-04-11
Add BigInteger support to NumberConverter

michelkaeser:


uklimaschewski:
Good idea, but there's some formatting problems with the code, see build details above.
Documentation at least in datataypes.md needs also to be extended.

PR #468, opened 2024-04-25
Add new function AVERAGE

oswaldobapvicjr:
This PR adds the function `AVERAGE(value, ...)` to calculate the arithmetic mean of the specified numbers.


uklimaschewski:
Thank you very much!

PR #471, opened 2024-05-08
EvaluationValue optimization

oswaldobapvicjr:
This PR optimizes the existing method `EvaluationValue.nullValue()` to return a shared, pre-built `EvaluationValue` instance (and not create a new object at each call).

Since it's an immutable object, it is safe to reuse the same instance without issues (similar to `Collections.emptyList()`, or `BigDecimal.ZERO`, for example).

By reusing the same object, we avoid GC overhead.

### DIFF
--- a/src/main/java/com/ezylang/evalex/Expression.java
+++ b/src/main/java/com/ezylang/evalex/Expression.java
@@ -367,7 +367,7 @@ public class Expression {
    * @return An {@link EvaluationValue} of the detected type and value.
    */
   public EvaluationValue convertValue(Object value) {
-    return new EvaluationValue(value, configuration);
+    return EvaluationValue.of(value, configuration);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


melontini:
I think booleans should be constant too, since they are only 2 possible states.

oswaldobapvicjr:
I found something strange in the constructor `EvaluationValue(Object, ExpressionConfiguration)` too:

```` java
  public EvaluationValue(Object value, ExpressionConfiguration configuration) {

    EvaluationValue converted =
        configuration.getEvaluationValueConverter().convertObject(value, configuration);

    this.value = converted.getValue();
    this.dataType = converted.getDataType();
  }
````

Here, the `converted` is already an `EvaluationValue`, why duplicate it into a new object?
Should we act on it too, @uklimaschewski ?

uklimaschewski:
Thank you very much

PR #472, opened 2024-05-08
Remove warnings from JUnit classes

oswaldobapvicjr:
This PR removes warnings from JUnit classes that used to extend HashMap to create test maps (old-fashioned). I am replacing these occurrences with a more modern option, introduced by JDK 9: `Map.of(K, V, ...)`.


uklimaschewski:
Thanks alot!

PR #482, opened 2024-06-21
New generic DataType.BINARY

oswaldobapvicjr:
This PR introduces a new `DataType.BINARY` to allow generic objects to be evaluated by special functions.

A new dedicated Converter -- `BinaryConverter` -- was created and added to the last position in the chain of converters in `DefaultEvaluationValueConverter`, allowing to convert any object that wasn't converted before, as per Discussion #481 (Working with XML Documents).

The existing behavior for objects that any converter could not convert was to throw an exception. I decided to preserve this behavior and let the new `BinaryConverter` be configurable in the `ExpressionConfiguration` so that the exception will not be thrown when the binary data type is enabled. I decided to let the initial state "false" for compatibility.

The `BinaryConverter` then reads the current configuration to determine whether or not to "convert" the object. To make that possible, I added the `Expressionconfiguration` as a new parameter to `ConverterIfc.canConvert(...)`.

I finally added some JUnits for full coverage according to the repository rules.

Documentation files were also updated accordingly.

### DIFF
--- a/docs/concepts/datatypes.md
+++ b/docs/concepts/datatypes.md
@@ -19,7 +19,10 @@ EvalEx supports the following data types:
 | ARRAY           | java.util.List                    |
 | STRUCTURE       | java.util.Map                     |
 | EXPRESSION_NODE | com.ezylang.evalex.parser.ASTNode |
+| BINARY[^1]      | java.lang.Object                  |
 | NULL            | null                              |
+
+[^1]: Since 3.3.0
 
 Data is stored in an _EvaluationValue_, which holds the value and the data type.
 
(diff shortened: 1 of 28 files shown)


uklimaschewski:
Thank you for the pull request and sorry for the late response!

I am a bit unhappy about changing the `ConverterIfc` interface, because this will break backward compatibility.
To do such things, a new major release version would be needed.

What do you think about just changing the `DefaultEvaluationConverter` instead?
Keep the list of converters, and after the loop, if no matching converter is found, either throw an exception or convert to binary if it is allowed:
```java
//...
    for (ConverterIfc converter : converters) {
      if (converter.canConvert(object)) {
        return converter.convert(object, configuration);
      }
    }

    if (configuration.isBinaryAllowed()) {
        return EvaluationValue.binaryValue(object);
    }

    throw new IllegalArgumentException(
        "Unsupported data type '" + object.getClass().getName() + "'");
```

oswaldobapvicjr:
New commit available

uklimaschewski (approved):

PR #513, opened 2024-10-24
New function STR_SPLIT

oswaldobapvicjr:
This push introduces new string functions:
- `STR_SUBSTRING_AFTER(string, separator)`
- `STR_SUBSTRING_BEFORE(string, separator)`

Examples:
- `STR_SUBSTRING_AFTER("file.json", ".")`
  - returns `"json"`
- `STR_SUBSTRING_BEFORE("2024/10/23", "/")`
  - returns `"2024"`


uklimaschewski:
Why not defining a single method `STR_SPLIT(string, separator)`? It could return an array of strings (similar to `String.split(String regex)`).

Using `STR_SPLIT("file.json", ".")[1]` would return `"json"`.
Using `STR_SPLIT("2024/10/23", "/")[0]` would return `"2024"`.

You would even have much more possibilities.

I am just unsure of using a regex or just a simple string as separator is better.

oswaldobapvicjr:
The functions `STR_SUBSTRING_AFTER` and `STR_SUBSTRING_BEFORE` were inspired by those provided by leading tools such as [XPath (W3C)](https://www.w3.org/TR/1999/REC-xpath-19991116/#section-String-Functions), [Microsoft Excel](https://support.microsoft.com/en-us/office/textafter-function-c8db2546-5b51-416a-9690-c7e6722e90b4), and [MuleSoft by Salesforce](https://docs.mulesoft.com/dataweave/latest/dw-strings-functions-substringafter) so I wonder it could be good here too, for simplicity and assimilation. However, I agree that the same result could be obtained using STR_SPLIT.

Regarding the doubt whether to use a regex or a separator, why not implement both? ;)

- `STR_SPLIT("file.json", ".")` would return `"json"` (simple and direct for everyone)
- `STR_SPLIT_REGEX("file.json", "\\.")` would also return `"json"`, but since it's regex-based, it offers more possibilities for users that are familiar with regex.

uklimaschewski:
Sorry for the late answer, this fell totally off my table.
I think having a `STR_SPLIT` and `STR_SPLIT_REGEX`is a good idea. I'd prefer these instead of having `STR_SUBSTRING_AFTER`and `STR_SUBSTRING_BEFORE`.

If you modify your PR accordingly, or open a new one I'd be happy to merge it.

oswaldobapvicjr:
I've just implemented the function `STR_SPLIT(string, separator)` but I'm now unsure about the need for a 2nd function (`STR_SPLIT_REGEX`), so I wonder if we may implement it in the future if we find a good use case for it. What do you think?

uklimaschewski:
Thank you very much! Adding the `STR_SPLIT_REGEX` function can be done anytime. I'll wait for a release a little while, mybe I or someone finds some time to add it - shouldn't be too much effort.

Issue #253, opened 2021-04-11
Anyway to make trigonometric functions implicit?


walsus:
I have 2 questions:

1- Can I make this expression acceptable ?   **6SIN(8)**    instead of **6xSIN(8)**   ?
2- Can I let **Sin(8**   be automatically evaluated when at the end of the expression ?without the need to add a closing parenthesis ?

uklimaschewski:
1- This would require some changes in the tokenizer, I guess. Depending on how many cases you want to support, this could be rather simple or quite complex.

2- You could do this outside EvalEx. Count the number of parentheses in the string. Then add the missing parentheses at the end, if there is a mismatch.

RobertZenz:
See #64 and #123 for pevious discussions.

uklimaschewski:
Oh, thanks @RobertZenz, I totally forgot about #123

Issue #64, opened 2016-04-06
Support for implicit multiplication


FlashTek:
Hello,
are there any plans to support implicit multiplications, so that `5SIN(2)` will be interpreted as `5*SIN(2)`? (Or is this even already possible and I am just unable to find this functionality?)

Edit: I tried to solve this issue by modifing the `shuntingYard` method by adding addition if-statements like this
    
    if (isNumber(token)) {
          if (previousToken != null && ")".equals(previousToken))
          {
               stack.push("×");
          }
         ...
    } else if (functions.containsKey(token.toUpperCase(Locale.ROOT))) {
         if (previousToken != null && isNumber(previousToken))
         {
                stack.push("×");
         }
     ....

This seems to work after a few test, but I am not completely sure about it. Are there any opinions?

uklimaschewski:
I am not very happy with this kind of syntax, it has the potential to open a whole new can of worms.
E.g. What about `A5BSIN(2)` - is this a function named `A5BSIN` or an expression `A*5*SIN(2)` ?

FlashTek:
Hm, that's a true problem. But are there any plans to support these multiplications? Because they are somehow very often used by users...

PR #534, opened 2025-07-14
Add implicit number-function multiplication

JasperLorelai:
As it stands, a token is only recognised as an identifier (usually a function) if it starts with a character or `_` (see [this](https://github.com/ezylang/EvalEx/blob/d6af50adf282752d5a7091d6895be1a6175a1e80/src/main/java/com/ezylang/evalex/parser/Tokenizer.java#L166-L169) and [this](https://github.com/ezylang/EvalEx/blob/d6af50adf282752d5a7091d6895be1a6175a1e80/src/main/java/com/ezylang/evalex/parser/Tokenizer.java#L582)). That allows us to add implicit multiplication, such as `2abs(x)`. This was proposed in some issues like #253 and #64, the latest of which opens up this concern:
> I am not very happy with this kind of syntax, it has the potential to open a whole new can of worms. E.g. What about `A5BSIN(2)` - is this a function named `A5BSIN` or an expression `A*5*SIN(2)` ?

`A5BSIN` should technically always be considered a function, just like `atan2(` is. This PR adds previously proposed number-to-function implicit multiplication.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ EvalEx is a handy expression evaluator for Java, that allows to parse and evalua
 - Custom functions and operators can be added.
 - Functions can be defined with a variable number of arguments (see MIN, MAX and SUM functions).
 - Supports hexadecimal and scientific notations of numbers.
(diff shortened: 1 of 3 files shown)


uklimaschewski:
Thanks for that extension

PR #537, opened 2025-07-26
Get available functions in the dictionary

f-lopes:
Add the possibility to get available functions in the dictionary.
This could be useful for implementing auto-completion in a UI, for example.

### DIFF
--- a/src/test/java/com/ezylang/evalex/config/MapBasedFunctionDictionaryTest.java
+++ b/src/test/java/com/ezylang/evalex/config/MapBasedFunctionDictionaryTest.java
@@ -70,7 +70,7 @@ class MapBasedFunctionDictionaryTest {
     functionsToRegister.forEach(dictionary::addFunction);
 
     assertThat(dictionary.getAvailableFunctionNames())
-        .containsAnyElementsOf(functionsToRegister.keySet());
+        .containsExactlyInAnyOrderElementsOf(functionsToRegister.keySet());
   }
 
   @Test
@@ -101,7 +101,7 @@ class MapBasedFunctionDictionaryTest {
     functionsToRegister.forEach(dictionary::addFunction);
 
     assertThat(dictionary.getAvailableFunctions())
-        .containsExactlyElementsOf(functionsToRegister.values());
+        .containsExactlyInAnyOrderElementsOf(functionsToRegister.values());
   }
 
   @Test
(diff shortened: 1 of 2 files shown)


f-lopes:
Hi,

Thank you for your review.

1. I agree with your point of view and with your proposal to solve this incompatibility. I will add a default function, as suggested.
2. Sure, I can make the change to return a set of `FunctionIfc`, or rename the interface method `getAvailableFunctionNames`. I will probably return a set of `FunctionIfc` to stay consistent with the `getFunction` method.
3. I hesitated to add this functionality to the `OperatorDictionaryIfc` as well, so I can add this. 

Please let me know your thoughts before I make the necessary changes.

Thank you.

src/test/java/com/ezylang/evalex/config/MapBasedFunctionDictionaryTest.java
```diff
@@ -59,4 +58,66 @@ void testCaseInsensitivity() {
...
+    HashMap<String, FunctionIfc> functionsToRegister = new HashMap<>();
+    functionsToRegister.put("min", new MinFunction());
+    functionsToRegister.put("max", new MaxFunction());
+
+    @SuppressWarnings({"unchecked", "varargs"})
+    FunctionDictionaryIfc dictionary = MapBasedFunctionDictionary.ofFunctions();
+    functionsToRegister.forEach(dictionary::addFunction);
+
+    assertThat(dictionary.getAvailableFunctionNames())
+        .containsAnyElementsOf(functionsToRegister.keySet());
```
uklimaschewski, line 73: `containsExactlyElementsOf()` is the better choice, I think.

uklimaschewski (changes requested):
It looks like the default functions in the interfaces are missing in the tests, therefore the code coverage check fails.
Can you please add them, here is an example on how to best fix this, from my point of view:
```java
package com.ezylang.evalex.config;

import com.ezylang.evalex.functions.FunctionIfc;
import org.junit.jupiter.api.Test;

import static org.assertj.core.api.Assertions.assertThatThrownBy;

public class FunctionDictionaryIfcTest {

    FunctionDictionaryIfc defaultDictionary = new FunctionDictionaryIfc() {
        @Override
        public void addFunction(String functionName, FunctionIfc function) {
        }

        @Override
        public FunctionIfc getFunction(String functionName) {
            return null;
        }
    };

    @Test
    void testGetAvailableFunctionsIsUnsupported() {
        assertThatThrownBy(() -> defaultDictionary.getAvailableFunctions()).isInstanceOf(UnsupportedOperationException.class);
    }
}
```

uklimaschewski:
Sorry for the late review, looks good but see my comments in the review. I will merge this, when those are fixed.

uklimaschewski (approved):

Issue #555, opened 2026-01-20
LenientMode not respected when accessing structure fields


AranyaniSharma:
### Description
The `lenientMode` configuration option is not properly respected when accessing structure fields in expressions. When `lenientMode(true)` is configured, accessing a missing field in a structure or accessing a property on `NULL_VALUE` still throws an `EvaluationException` instead of gracefully returning `NULL_VALUE`.

### Environment
- **EvalEx Version**: 3.6.0

### Expected Behavior
When `lenientMode` is enabled:
- Accessing a missing field in a structure should return `NULL_VALUE` instead of throwing an exception
- Accessing a property on `NULL_VALUE` should return `NULL_VALUE` instead of throwing an exception

### Actual Behavior
The `evaluateStructureSeparator()` method in `Expression.java` always throws exceptions for:
1. Missing fields: `Field 'xyz' not found in structure`
2. Property access on null: `Unsupported data types in operation`

Even when `configuration.isLenientMode()` returns `true`.

### Example Code to Reproduce

ExpressionConfiguration config = ExpressionConfiguration.builder()
    .lenientMode(true)
    .build();

Map<String, Object> data = new HashMap<>();

Expression expression = new Expression("user.name", config);
EvaluationValue result = expression.evaluate(); // Throws exception instead of returning NULL

AranyaniSharma:
### Root Cause
The `evaluateStructureSeparator()` method in `Expression.java` (lines ~220-235) does not check `configuration.isLenientMode()` before throwing exceptions.

### Impact
This prevents using EvalEx for null-safe navigation in expressions where nested data structures might have missing or null values, which is a common use case in:
- Data processing pipelines with optional/nullable fields
- JSON/REST API response processing
- Dynamic configuration evaluation
- Any scenario where graceful fallback is needed

### Proposed Fix
Modify the `evaluateStructureSeparator()` method in `Expression.java` to check `lenientMode` before throwing exceptions:


private EvaluationValue evaluateStructureSeparator(ASTNode startNode, int depth)
    throws EvaluationException {
  EvaluationValue structure = evaluateSubtree(startNode.getParameters().get(0), depth + 1);
  Token nameToken = startNode.getParameters().get(1).getToken();
  String name = nameToken.getValue();

  if (structure.isStructureValue()) {
    if (!structure.getStructureValue().containsKey(name)) {
      // FIX: Check lenientMode before throwing exception for missing field
      if (configuration.isLenientMode()) {
[...]

PR #558, opened 2026-02-18
Fix issue #555: Lenient mode not respected when accessing structure fields

oswaldobapvicjr:
This solves the reported issue #555


oswaldobapvicjr:
@uklimaschewski, can you approve this fix and the next one (#559) as well, then release a patch version (3.6.1)?

uklimaschewski:
sorry for being late

PR #568, opened 2026-06-11
Add JPMS module descriptor and fix Lombok/Java 11 build compatibility

Beowolve:
- Added `module-info.java` with module `com.ezylang.evalex`
  - Added `requires static lombok` for compile-time Lombok usage
  - Exported the library’s public API packages, including `com.ezylang.evalex.functions.string.util`
  - Opened packages for reflective access as a temporary compatibility measure (tests/frameworks)

- Fixed Lombok compilation on Java 11 by explicitly configuring annotation processing via `maven-compiler-plugin`

- Updated the Spotless `googleJavaFormat` version to `1.24.0` (the Spotless Maven plugin version itself is unchanged)

The broad `opens` directives are currently kept for compatibility and can be narrowed in a follow-up once reflective usage is reduced.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
     <groupId>com.ezylang</groupId>
     <artifactId>EvalEx</artifactId>
-    <version>3.7.0-SNAPSHOT</version>
+    <version>3.7.1-SNAPSHOT</version>
 
     <name>EvalEx</name>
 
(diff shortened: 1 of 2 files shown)


uklimaschewski:
Thanks for the PR. The general direction makes sense, especially adding a JPMS module descriptor and making Lombok annotation processing explicit. However, I would not merge this as-is.

Please remove the project version bump from this PR, as version changes should be handled separately during release preparation.

Also, the statement “exported all packages” is not fully correct. At least com.ezylang.evalex.functions.string.util exists but is not exported or opened. Since JPMS does not include subpackages automatically, this could still introduce a breaking change for modular users.

Finally, opening all packages for reflection is acceptable as a temporary compatibility measure, but we should either document this clearly in module-info.java or restrict it where possible in a follow-up. The current module descriptor effectively exports and opens almost the whole library, which is compatible but not a strong module boundary.

Please also adjust the PR description: the Spotless plugin itself is not updated to 1.24.0; the googleJavaFormat version is.

Beowolve:
Thanks for the detailed review — I’ve addressed the requested changes.

- Removed the project version bump from this PR.
- Updated the module descriptor wording to clarify that broad `opens` is a temporary compatibility measure.
- Synced with upstream and added `exports`/`opens` for `com.ezylang.evalex.functions.string.util` so JPMS subpackage behavior is handled explicitly.
- Corrected the PR description: this PR updates `googleJavaFormat` to 1.24.0 in Spotless config; it does not update the Spotless Maven plugin itself.

Please take another look when you have a moment.

uklimaschewski:
Thanks!

PR #571, opened 2026-06-27
Fix CWE-400: Add upper limit to FACT function to prevent DoS attacks

### LINKED ISSUE #570: [Security] DoS vulnerabilities in STR_MATCHES and FACT function
Hi,

I'm a security researcher and I've found two denial-of-service vulnerabilities in EvalEx 3.6.2:

1. **CWE-1333 (ReDoS):** The `STR_MATCHES` function passes user-controlled regex directly to `String.matches()`, allowing catastrophic backtracking with a crafted pattern.
2. **CWE-400 (Uncontrolled Resource Consumption):** The `FACT` function has no upper bound on input, enabling CPU exhaustion via large values.

Both are confirmed with a working PoC on the latest version.

I noticed that GitHub Security Advisories (private reporting) is not enabled for this repository. Could you let me know the preferred channel for sharing the full technical details and PoC?

I follow responsible disclosure practices and am happy to coordinate on a timeline that works for you.

Thank you!

oswaldobapvicjr:
This commit addresses the security vulnerability reported in issue #570 by implementing a maximum input limit for the FACT function, defined by the existing configuration property 'maxRecursionDepth' (initial/default value: 2000)

Fixes #570

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -24,6 +24,7 @@ ExpressionConfiguration configuration=ExpressionConfiguration.builder()
         .lenientMode(false)
         .locale(Locale.getDefault())
         .mathContext(ExpressionConfiguration.DEFAULT_MATH_CONTEXT)
+        .maxRecursionDepth(2000)
         .operatorDictionary(ExpressionConfiguration.StandardOperatorsDictionary)
         .powerOfPrecedence(OperatorIfc.OPERATOR_PRECEDENCE_POWER)
         .stripTrailingZeros(true)
@@ -145,6 +146,12 @@ The math context is used throughout all operations and functions. The default ha
 and a rounding mode of _HALF_EVEN_.
 
 See chapter [Precision, Scale and Rounding](../concepts/rounding.html) for details.
+
+### Maximum Recursion Depth[^2]
+
+Specifies the maximum number of nested call frames allowed during recursive operations (such as resolving
+nested expressions or calculating factorials). This parameter protects the system against stack overflow
+errors, CPU exhaustion, and Denial of Service (DoS) attacks. The default value is 2000.
 
 ### Operator Dictionary
 
--- a/src/main/java/com/ezylang/evalex/functions/basic/FactFunction.java
+++ b/src/main/java/com/ezylang/evalex/functions/basic/FactFunction.java
@@ -28,12 +28,11 @@ import java.math.BigDecimal;
 public class FactFunction extends AbstractFunction {
 
   /**
-   * Maximum allowed input value for factorial calculation.
-   * Set to 170 to align with industry-standard limits in Windows Calculator, macOS Calculator,
-   * Microsoft Excel, Google Sheets, and LibreOffice Calc.
-   * This prevents uncontrolled resource consumption (CWE-400) while maintaining compatibility
-   * with user expectations from widely-used tools.
-   * See: https://github.com/ezylang/EvalEx/issues/570
+   * Maximum allowed input value for factorial calculation, aligned with industry-standard upper
+   * limits. This prevents uncontrolled resource consumption (CWE-400) while maintaining
+   * compatibility with user expectations from widely-used tools.
+   *
+   * @see https://github.com/ezylang/EvalEx/issues/570
    */
   private static final int MAX_FACTORIAL_INPUT = 170;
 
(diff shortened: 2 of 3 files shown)


oswaldobapvicjr:
@uklimaschewski,
I created this PR in response to the potential DoS vulnerability identified in #570. 

Limiting the input to prevent a Denial of Service (DoS) via the `fact()` function is definitely the right move. However, instead of hardcoding a fixed limit like 170 (my initial proposal), I think it makes more sense to reuse our existing `maxRecursionDepth`configuration.

Since EvalEx relies on `BigDecimal` for high-precision math, users might legitimately want to calculate larger factorials if their environment allows it. By leveraging `maxRecursionDepth`, we can accomplish a few things:

1. **Maintain consistency:** Reuse an existing security mechanism designed to prevent stack overflows and resource exhaustion.

2. **Keep it configurable:** Give developers the control to tighten or loosen the restriction based on their specific use case, rather than forcing a hardcoded limit.

3. **Secure by default, flexible by choice:** The library should be secure out of the box for the vast majority of users. However, for those who use EvalEx in secure, offline environments and genuinely rely on heavy computations, tying this to `maxRecursionDepth` allows them to raise or disable the limit to suit their needs easily.

4. **Conceptually Aligned:** Even though the factorial function is not implemented recursively, mathematically and computationally, it represents an O(n) linear growth in resource consumption, just like recursion depth.

5. **Easier Maintenance**: For you as a maintainer, it means one less configuration path to document, test, and support down the road.

Let me know if you also agree to reuse the config here, and I'll be happy to submit a new commit with the change.

dyingman1:
Thank you for the detailed feedback and for acknowledging the issue. 

I completely agree with your approach. Reusing `maxRecursionDepth` makes a lot of sense conceptually since both deal with linear resource consumption, and it provides great flexibility for users in secure environments while keeping the library secure out of the box. Please go ahead with the new commit using the existing configuration.

Additionally, since this DoS vulnerability has been confirmed and is being patched, I would like to request a CVE ID for this issue to help track it and ensure users are aware of the security update. 

Do you plan to open a GitHub Security Advisory (GHSA) to publish this and request a CVE? If you prefer, I can also proceed with filing the CVE request directly. Please let me know your thoughts on how we should handle the CVE assignment.

Thank you again for your time and great collaboration!

oswaldobapvicjr:
Thanks for the suggestion, @dyingman1. I agree that opening a GHSA and requesting the CVE through it sounds like the right path to ensure awareness, so users can update to the latest secure version when it's ready.

If you'd like to go ahead and open the GHSA / request the CVE ID yourself, feel free to do so. We'd just ask that you keep the advisory in draft (private) until the fix is released, so users have a patched version available at the time of public disclosure.

In the meantime, the fix itself is on track — I'll make sure the commit is ready for review so we don't lose momentum on remediation.

Would that work for you?

Thanks again for the collaboration — really appreciate how smoothly this has gone.

dyingman1:
Thanks for the follow-up! 

I just checked the repository, but it looks like the "Report a vulnerability" feature (Private Vulnerability Reporting) isn't enabled at the moment, which prevents me from initiating the GHSA from my side.

Could you please either enable "Private vulnerability reporting" in the repository's security settings, or simply create the draft GHSA from your end and add me (@dyingman1) as a collaborator? Either way works perfectly for me.

Thanks again for your support and for keeping this moving so smoothly!

oswaldobapvicjr:
Hello, I just submitted a new commit that moves from hardcoded to a configurable upper limit via 'maxRecursionDepth'. JUnit tests validate the fix. The property was also documented (for some reason, it was missing in the doc files).

oswaldobapvicjr:
> Thanks for the follow-up!
> 
> I just checked the repository, but it looks like the "Report a vulnerability" feature (Private Vulnerability Reporting) isn't enabled at the moment, which prevents me from initiating the GHSA from my side.
> 
> Could you please either enable "Private vulnerability reporting" in the repository's security settings, or simply create the draft GHSA from your end and add me (@dyingman1) as a collaborator? Either way works perfectly for me.
> 
> Thanks again for your support and for keeping this moving so smoothly!

I'm just a contributor, and I don't have access to the repository settings. @uklimaschewski is the project administrator.

@uklimaschewski, can you check this?

oswaldobapvicjr:
The PR is ready for review.

PR #572, opened 2026-07-01
Fix CWE-1333: Add timeout to STR_MATCHES to prevent ReDoS attacks

oswaldobapvicjr:
This PR addresses the security vulnerability reported in issue #570 by implementing a timeout on the RegEx matching operation inside the STR_MATCHES function. The timeout is configurable via a new configuration parameter 'regexTimeoutMillis' (initial/default value: 100ms).

### DIFF
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -26,6 +26,7 @@ ExpressionConfiguration configuration=ExpressionConfiguration.builder()
         .mathContext(ExpressionConfiguration.DEFAULT_MATH_CONTEXT)
         .operatorDictionary(ExpressionConfiguration.StandardOperatorsDictionary)
         .powerOfPrecedence(OperatorIfc.OPERATOR_PRECEDENCE_POWER)
+        .regexTimeoutMillis(100)
         .stripTrailingZeros(true)
         .structuresAllowed(true)
         .binaryAllowed(false)
@@ -175,6 +176,12 @@ ExpressionConfiguration configuration=ExpressionConfiguration.builder()
 // will now result in -4, instead of 4:
         Expression expression=new Expression("-2^2",configuration);
 ```
+
+### RegEx Timeout in Milliseconds
+
+Defines the maximum execution time allowed for a single regular expression matching operation.
+This parameter serves as a critical security boundary to stop Catastrophic Backtracking, which can
+freeze the CPU and lead to Denial of Service (DoS) attacks. The default value is 100 milliseconds.
 
 ### Single Quote String Literals
 
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -177,11 +177,13 @@ ExpressionConfiguration configuration=ExpressionConfiguration.builder()
         Expression expression=new Expression("-2^2",configuration);
 ```
 
-### RegEx Timeout in Milliseconds
+### RegEx Timeout in Milliseconds[^4]
 
 Defines the maximum execution time allowed for a single regular expression matching operation.
 This parameter serves as a critical security boundary to stop Catastrophic Backtracking, which can
 freeze the CPU and lead to Denial of Service (DoS) attacks. The default value is 100 milliseconds.
+
+[^4]: Since 3.6.3
 
 ### Single Quote String Literals
 
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -187,8 +187,12 @@ ExpressionConfiguration configuration=ExpressionConfiguration.builder()
 ### RegEx Timeout in Milliseconds[^4]
 
 Defines the maximum execution time allowed for a single regular expression matching operation.
-This parameter serves as a critical security boundary to stop Catastrophic Backtracking, which can
-freeze the CPU and lead to Denial of Service (DoS) attacks. The default value is 100 milliseconds.
+
+This parameter serves as a critical operational security breaker to bound Catastrophic Backtracking,
+which can otherwise consume excessive CPU resources and lead to Denial of Service (DoS) attacks.
+
+Setting this value to 0 or any negative number disables the timeout enforcement completely, allowing
+regular expression evaluations to run without any time constraints. The default value is 100 milliseconds.
 
 [^4]: Since 3.6.3
 
--- a/docs/configuration/configuration.md
+++ b/docs/configuration/configuration.md
@@ -186,10 +186,8 @@ ExpressionConfiguration configuration=ExpressionConfiguration.builder()
 
 ### RegEx Timeout in Milliseconds[^4]
 
-Defines the maximum execution time allowed for a single regular expression matching operation.
-
-This parameter serves as a critical operational security breaker to bound Catastrophic Backtracking,
-which can otherwise consume excessive CPU resources and lead to Denial of Service (DoS) attacks.
+This parameter limits the maximum runtime of a single regular expression matching operation, helping to
+mitigate ReDoS risks caused by catastrophic backtracking.
 
 Setting this value to 0 or any negative number disables the timeout enforcement completely, allowing
 regular expression evaluations to run without any time constraints. The default value is 100 milliseconds.
--- a/src/main/java/com/ezylang/evalex/config/ExpressionConfiguration.java
+++ b/src/main/java/com/ezylang/evalex/config/ExpressionConfiguration.java
@@ -168,6 +168,9 @@ public class ExpressionConfiguration {
   /** The default maximum depth for recursion is 2000 levels. */
   public static final int DEFAULT_MAX_RECURSION_DEPTH = 2_000;
 
+  /** The default RegEx timeout is 100 milliseconds. */
+  public static final long DEFAULT_REGEX_TIMEOUT_MILLIS = 100L;
+
   /**
    * The default date time formatters used when parsing a date string. Each format will be tried and
    * the first matching will be used.
@@ -388,6 +391,12 @@ public class ExpressionConfiguration {
 
   /** The maximum recursion depth allowed for nested expressions. */
   @Builder.Default private final int maxRecursionDepth = DEFAULT_MAX_RECURSION_DEPTH;
+
+  /**
+   * Timeout in milliseconds for RegEx matching execution. Prevents catastrophic backtracking in
+   * RegEx patterns from consuming excessive CPU resources.
+   */
+  @Builder.Default private final long regexTimeoutMillis = DEFAULT_REGEX_TIMEOUT_MILLIS;
 
   /**
    * The date-time formatters. When parsing, each format will be tried and the first matching will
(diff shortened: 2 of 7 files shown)


oswaldobapvicjr:
Hi @uklimaschewski, the build is failing because the minimum coverage requirement for new PRs isn't met.

Coverage Justification: The uncovered blocks are defensive catch clauses for InterruptedException and TimeoutException wrapping a Future operation. Forcing the simulation of these asynchronous edge cases via JUnit requires brittle, multi-threaded test boilerplate that introduces environment-dependent flaky tests into our CI/CD pipeline. The code already adheres to standard concurrency best practices (safely restoring the thread's interrupt flag). Therefore, I request a coverage gate exemption for these specific defensive lines to avoid maintaining high-overhead, low-value tests.

uklimaschewski:
Thank you very much for taking the time to work on this issue. I really appreciate your contribution.

After reviewing the approach, I don't think we should merge it in its current form. The reason is not the implementation itself, but a limitation of `java.util.regex`.

The timeout prevents the caller from waiting forever, but it does not guarantee that the regex matching actually stops. As far as I know, Java's regex engine does not reliably react to thread interruption. This means that a regex with catastrophic backtracking may continue using CPU even after the timeout has expired. With enough malicious requests, the shared thread pool could eventually become full.

Because of this, I don't think this completely fixes the reported ReDoS vulnerability. I'm also worried that the documentation could give the impression that catastrophic backtracking is prevented, while the timeout only limits how long the caller waits.

For these reasons, I don't feel comfortable merging this as the official fix. I think we should instead look at a solution that removes the root cause, for example by using a linear-time regex engine such as RE2/J, even if this means supporting a smaller set of regex features.

Thank you again for your contribution. I think the timeout approach could still be useful as an optional extra layer of protection, but I don't think it should be presented as the main fix for this security issue.

oswaldobapvicjr:
Thanks for pointing that out @uklimaschewski. You raise a valid concern — `java.util.regex` doesn't cooperate with thread interruption, so a timeout only masks the symptom from the caller's perspective while the backtracking thread keeps burning CPU. I agree this isn't a complete fix.

So we need a different solution for STR_MATCHES. I'd consider using RE2/J for safer RegEx, since it guarantees linear-time matching by construction — which eliminates the root cause rather than working around it. But I know your preference about avoiding dependencies here inside EvalEx.

That said, I'm afraid we cannot simply disable or remove `STR_MATCHES`, as it's been available since version 3.4.0. I work on a commercial product that uses EvalEx to support a few use cases (including STR_MATCHES against simple patterns), and it would be disappointing if the function gets removed. I'd expect other users to be in the same situation. Beyond that, a published CVE against EvalEx without an available fix forces downstream consumers into a difficult position — we'd need to carry security exemptions, justify the risk to auditors. That's a burden I'd rather avoid if we can solve it properly here.

I also checked whether newer JDK versions would help — they don't. `java.util.regex` is still a backtracking NFA engine even in JDK 25. There's a draft JEP (8260688) for predictable regex performance, but it's not targeted to any release.

I see your reply about moving the function to a separate module to keep the core dependency-free (like EvalEx-big-math). That's interesting. However, since STR_MATCHES has been part of the default configuration, and removing it from core would break existing users silently at runtime (per semver, that would require a major version bump to 4.0, which seems disproportionate for what could be solved with a single lightweight dependency on RE2/J).

PS: I would like to dig a little deeper and test the approach discussed here (zero extra threads, immediate termination): https://stackoverflow.com/questions/910740/cancelling-a-long-running-regex-match

oswaldobapvicjr:
@uklimaschewski, Just submitted a new variation that avoids using additional threads and has no additional dependencies. Tested. Please approve.

uklimaschewski:
Thank you very much for the updated implementation. I like this approach much better than the previous thread-based solution.

Wrapping the input in a custom `CharSequence` is a clever idea. It avoids the problems of thread interruption and limits the maximum time spent evaluating a regular expression. While it does not change the behavior of Java's regex engine itself, it significantly mitigates the ReDoS risk by preventing unbounded catastrophic backtracking.

Before merging, I would like to ask for a few additional improvements:

1. **Start the timeout when matching starts**

   The timeout currently starts when the `Matcher` is created, not when matching starts. For the current `STR_MATCHES` usage this may not matter, but for a general utility method returning a `Matcher` this is surprising.

   Either the utility should directly execute the match operation, or the timeout should start as close as possible to the actual matching call.

2. **Share the timeout in `subSequence()`**

   `subSequence()` should reuse the original timeout/deadline instead of creating a new timeout.

   Otherwise the effective timeout may be extended if the regex engine creates subsequences during matching.

3. **Use `System.nanoTime()`**

   Since this is measuring elapsed time rather than wall-clock time, `System.nanoTime()` would be better than `System.currentTimeMillis()`.

4. **Handle invalid regular expressions**

   `PatternSyntaxException` should be handled explicitly and converted into an appropriate `EvaluationException`.

5. **Adjust documentation wording**

   I would describe this as a timeout-based protection that limits the runtime of regular expression evaluation.

   I would avoid wording that sounds like Java regex itself becomes safe or that catastrophic backtracking is prevented completely. The backtracking may still happen, but the timeout prevents it from running without bounds.

6. **Add a compatibility switch**

   The timeout-protected implementation should become the new default for `STR_MATCHES`.

   However, I would like to keep the existing direct `String.matches()` behavior available through a configuration option for users who need full compatibility with the previous behavior. Maybe we can use a timeout of 0 to disable the timeout check and use the old direct `String.matches()` functionality.

7. **Deprecate the legacy behavior**

   The legacy direct `String.matches()` behavior should be marked as deprecated since `4.0.0`.

   This should be documented both in the source code and in the user documentation.

   The documentation should clearly recommend the new timeout-protected implementation as the default and safer option.

I think this gives us a good balance:

- existing applications can keep the old behavior if they really need it
- new applications get better protection by default
- EvalEx avoids adding an external regex dependency
- the migration path is clear and documented

Thank you again for the contribution. I think this is a very good direction, and after these changes I would be happy to merge it.

oswaldobapvicjr:
@uklimaschewski, thanks again for the detailed feedback! I have refactored the implementation to address your concerns and align it with EvalEx's architecture. Here is how each point from your review has been handled:

1. **Start the timeout when matching starts (Resolved via API Design):**
You mentioned that calculating the timeout during Matcher creation could be problematic if the Matcher is left idle before execution. To address this permanently, I changed `createMatcher()` to a private helper method and exposed only the `matches(...)` utility. Because the Matcher is now instantiated and immediately executed (matcher.matches()) within the same method scope, the deadline is effectively calculated right when matching starts. This eliminates the risk of an idle `Matcher` burning through its timeout budget.

2. **Share the timeout in subSequence() (Fixed):**
The `TimeoutRegexCharSequence` constructor now accepts an absolute deadline timestamp (timeoutNano) instead of a relative duration. When `subSequence()` is called, it passes the same timeout forward to the new instance, ensuring that the expiration window is shared and cannot be extended or reset by backtracking routines.

3. **Use System.nanoTime() (Fixed):**
Switched entirely from wall-clock milliseconds to monotonic nanoseconds (but kept the configuration in milliseconds for simplicity).

5. **Handle invalid regular expressions (Fixed):**
`PatternSyntaxException` is now explicitly caught inside `createMatcher()` and cleanly wrapped into an EvalEx `EvaluationException`, passing along the current Token for accurate error reporting and positioning.

6. **Adjust documentation wording (Fixed):**
Updated the Javadocs to characterize this utility as a timeout-based operational circuit breaker explicitly. It clarifies that catastrophic backtracking can still occur algorithmically, but its runtime impact is strictly bounded.

7. **Add a compatibility switch (Fixed):**
The utility now dynamically fetches the timeout configuration via `expression.getConfiguration().getRegexTimeoutMillis()`. If a user configures this value to 0 or a negative integer, the code effectively disables the timeout protection.

8. **Deprecate the legacy behavior (Acknowledged):**
I have tagged the new Javadocs with `@since 3.6.3` based on the current version line. Since timeout can be turned off, I assume it's fine to introduce it in a patch release. Regarding deprecation of legacy pathways, I will defer to your preference on how you'd like to handle the long-term strategy for the upcoming releases. Feel free to adjust those boundaries and target versions during the merge!

Please take a look at the updated code — I'd love to hear your thoughts!

docs/configuration/configuration.md
```diff
@@ -183,6 +184,18 @@ ExpressionConfiguration configuration=ExpressionConfiguration.builder()
         Expression expression=new Expression("-2^2",configuration);
 ```
 
+### RegEx Timeout in Milliseconds[^4]
+
+Defines the maximum execution time allowed for a single regular expression matching operation.
+
+This parameter serves as a critical operational security breaker to bound Catastrophic Backtracking,
```
uklimaschewski: I would prefer a little less strong wording here:
This parameter limits the maximum runtime of a single regular expression matching operation.
It helps mitigate ReDoS risks caused by catastrophic backtracking.

src/test/java/com/ezylang/evalex/functions/string/util/RegularExpressionUtilsTest.java
```diff
@@ -0,0 +1,88 @@
+/*
```
uklimaschewski, line 1: Please have a look at the [SonarQube findings](https://github.com/ezylang/EvalEx/pull/572/checks?check_run_id=85505891964)

They are also commented here  in that file.

src/main/java/com/ezylang/evalex/config/ExpressionConfiguration.java
```diff
@@ -389,6 +392,20 @@ public class ExpressionConfiguration {
   /** The maximum recursion depth allowed for nested expressions. */
   @Builder.Default private final int maxRecursionDepth = DEFAULT_MAX_RECURSION_DEPTH;
 
+  /**
+   * The maximum execution time allowed for a single regular expression matching operation.
+   *
+   * <p>This parameter serves as a critical operational security breaker to bound Catastrophic
+   * Backtracking, which can otherwise consume excessive CPU resources and lead to Denial of Service
+   * (DoS) attacks.
```
uklimaschewski: Same wording adjustment as in the `configuration.md` here, too.

uklimaschewski (changes requested):
Thanks, that look very good. Only a few small corrections. Wording of the documentation and 3 Sonar findings.

oswaldobapvicjr (commented):
Comments addressed. Submitting new commit.

uklimaschewski (commented):
Always have a look at Sonar: https://sonarcloud.io/project/issues?id=ezylang_EvalEx&pullRequest=572&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

Can you please fix them?

oswaldobapvicjr:
> Always have a look at Sonar: https://sonarcloud.io/project/issues?id=ezylang_EvalEx&pullRequest=572&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true
> 
> Can you please fix them?

Just solved with a new commit.

uklimaschewski:
Thanks. I'll merge tomorrow and then create a new release.

uklimaschewski (approved):
Thank you very much, this looks very good now.

PR #578, opened 2026-08-11
New function DT_TRUNCATE

oswaldobapvicjr:
## Enhancement
This PR introduces an essential data-engineering feature: the `DT_TRUNCATE` function. 

Truncating timestamps is a fundamental operation for normalizing temporal data, especially when generating business reports, grouping metrics by specific intervals (hourly, daily, monthly), or stripping time components for date-only comparisons.

### Examples

| Expression                                                                   | Result                 |
|------------------------------------------------------------------------------|------------------------|
| `DT_TRUNCATE(DT_DATE_PARSE("2026-08-08T14:30:45Z"), "m")`                    | `2026-08-08T14:30:00Z` |
| `DT_TRUNCATE(DT_DATE_PARSE("2026-08-08T14:30:45Z"), "H")`                    | `2026-08-08T14:00:00Z` |
| `DT_TRUNCATE(DT_DATE_PARSE("2026-08-08T14:30:45Z"), "d")`                    | `2026-08-08T00:00:00Z` |
| `DT_TRUNCATE(DT_DATE_PARSE("2026-08-08T14:30:45Z"), "M")`                    | `2026-08-01T00:00:00Z` |
| `DT_TRUNCATE(DT_DATE_PARSE("2026-08-08T14:30:45Z"), "y")`                    | `2026-01-01T00:00:00Z` |

### Market Alignment & Justification
Major data processing and enterprise systems offer dedicated capabilities for this exact use case, highlighting its demand:
- **Relational Databases (Oracle, Snowflake, PostgreSQL):** Heavily rely on functions like `TRUNC(date, format)` or `DATE_TRUNC('unit', date)`.
- **Integration Engines (MuleSoft/DataWeave):** Lack a direct function, forcing developers into anti-patterns like manually calculating and subtracting periods (`now() - seconds(now().seconds)`) or reconstructing nested objects component by component.
- **Spreadsheets (Excel/Google Sheets):** Force users into long, nested combinations like `DATE(YEAR(A1), MONTH(A1), 1)` just to truncate to the start of a month.

By introducing `DT_TRUNCATE`, EvalEx bridges this gap, giving users a clean, high-performance, single-function alternative to nested or mathematical date logic.

### Architectural Design Choices
To ensure seamless integration with EvalEx's core philosophies, the function was designed with the following principles:
1. **Internal Consistency:** Instead of inventing a proprietary vocabulary for time units, it leverages the standard Java `DateTimeFormatter` characters (`y`, `M`, `d`, `H`, `m`, `s`). Users who already write masks for `DT_DATE_FORMAT` will immediately know how to use this function without a learning curve.
2. **Strict Case-Sensitivity:** By strictly preserving character casing, we cleanly separate calendar units from time fractions, completely avoiding the universal conflict between **M**onth (`"M"`) and **m**inute (`"m"`).
3. **Positional Parameter Flexibility:** Following EvalEx's native pattern for optional arguments (like `DT_DATE_PARSE`), parameters default sensibly (`timeUnit` defaults to day `"d"`, and `zoneId` defaults to the expression configuration). Bypassing the middle argument to supply a custom zone is gracefully handled via `NULL` (e.g., `DT_TRUNCATE(dt, NULL, "America/Chicago")`).
4. **TimeZone Aware Instant Conversion:** Since `EvaluationValue` holds times as UTC `Instant` objects, the truncation boundaries are safely evaluated from a local calendar perspective using `ZoneId` before wrapping back into an `Instant`, protecting the engine from CI/CD pipeline environment shifts.

## Changes Included
- **Core Feature:** `DateTimeTruncateFunction.java` implementation.
- **Test Coverage:** Added new tests into `DateTimeFunctionsTest.java`.
- **Documentation:** Updated the markdown documentation following the exact table structure, styling, and UTC examples found in sibling functions like `DT_DATE_FORMAT`.

### DIFF
--- a/docs/references/functions.md
+++ b/docs/references/functions.md
@@ -103,6 +103,7 @@ Available through the _ExpressionConfiguration.StandardFunctionsDictionary_ cons
 | [DT_DURATION_TO_MILLIS](#dt_duration_to_millis)     | Converts the given _DURATION_ to a milliseconds value.                                                                                                                                                                                                                      | 
 | [DT_NOW](#dt_now)                                   | Produces a new _DATE_TIME_ that represents the current moment in time.                                                                                                                                                                                                      |
 | [DT_TODAY](#dt_today)                               | Produces a new _DATE_TIME_ that represents the current date, at midnight (00:00). An optional time zone can be specified, e.g. "America/Sao_Paulo", or "GMT-03:00". If no zone ID is specified, the configured zone ID is used.                                             |
+| [DT_TRUNCATE](#dt_truncate)                         | Truncates the given _DATE_TIME_ value to a specific time unit. |
 
 ---
 
@@ -2479,3 +2480,46 @@ Use this when you need a reference to today’s date without the current time co
 \* Example considering UTC as system time zone
 
 🔝 [Back to Date Time Functions](#date-time-functions) | 🔝 [Back to top](#top)
+
+
+## DT_TRUNCATE
+
+*Since: 3.8.0*
+
+The `DT_TRUNCATE` function truncates a date-time value to a specific time unit, resetting smaller units to their start value, allowing developers to normalize timestamps for groupings, reports, or interval calculations. You can optionally supply a specific time unit and a target time zone. If omitted, the day ("d") unit and the system’s configured default time zone are used.
+
+### Syntax
+
+```
+DT_TRUNCATE(value [, timeUnit [, zoneId]]
+```
+
+### Parameters
+
+| Name     | Description                                                                                                                                                                 |
+|----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| value    | The `DATE_TIME` value to truncate.                                                                                                                                          |
+| timeUnit | *(Optional)* Case-sensitive formatting character. Default is `"d"`. Supported units: `"y"` (Year), `"M"` (Month), `"d"` (Day), `"H"` (Hour), `"m"` (Minute), `"s" (Second). |
+| zoneId   | *(Optional)* Time zone identifier string to use (e.g. `"UTC"`, `"America/Sao_Paulo"`). To skip `timeUnit` and pass a zone, use `NULL` for `timeUnit`.                       |
+
+
+### Examples
+
+These examples show how various combinations of units and zones affect the resulting date-time value.
+
+>**Note:** The function [DT_DATE_PARSE](#dt_date_parse) is being used to produce `DATE_TIME` values for the truncation operations.
+
+| Expression                                                                   | Result                 |
+|------------------------------------------------------------------------------|------------------------|
+| `DT_TRUNCATE(DT_DATE_PARSE("2026-08-08T14:30:45Z"))`                         | `2026-08-08T00:00:00Z` |
+| `DT_TRUNCATE(DT_DATE_PARSE("2026-08-08T14:30:45Z"), "m")`                    | `2026-08-08T14:30:00Z` |
+| `DT_TRUNCATE(DT_DATE_PARSE("2026-08-08T14:30:45Z"), "H")`                    | `2026-08-08T14:00:00Z` |
+| `DT_TRUNCATE(DT_DATE_PARSE("2026-08-08T14:30:45Z"), "d")`                    | `2026-08-08T00:00:00Z` |
+| `DT_TRUNCATE(DT_DATE_PARSE("2026-08-08T14:30:45Z"), "M")`                    | `2026-08-01T00:00:00Z` |
+| `DT_TRUNCATE(DT_DATE_PARSE("2026-08-08T14:30:45Z"), "y")`                    | `2026-01-01T00:00:00Z` |
+| `DT_TRUNCATE(DT_DATE_PARSE("2026-08-08T14:30:45Z"), "d", "America/Chicago")` | `2026-08-08T05:00:00Z` |
+
+\* Examples considering UTC as system time zone
+
+🔝 [Back to Date Time Functions](#date-time-functions) | 🔝 [Back to top](#top)
+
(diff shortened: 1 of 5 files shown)


src/main/java/com/ezylang/evalex/functions/datetime/DateTimeTruncateFunction.java
```diff
@@ -0,0 +1,156 @@
...
+      try {
+        zoneId = ZoneId.of(zoneString);
+      } catch (Exception e) {
+        throw new EvaluationException(
+            functionToken, "Time zone with id '" + zoneString + "' not found");
+      }
+    }
+
+    // Apply truncation
+    LocalDateTime localDateTime = LocalDateTime.ofInstant(instant, zoneId);
```
uklimaschewski: Converting the input `Instant` to `LocalDateTime` drops which offset was active during a daylight-saving overlap. For example, `2026-11-01T06:30:00Z` is the second `01:30` in `America/New_York`; truncating it to the hour through `LocalDateTime` and calling `atZone()` selects the earlier `-04:00` offset and returns `05:00Z`, whereas the correct boundary for that occurrence is `06:00Z`. Please keep the value as a `ZonedDateTime` and truncate it directly so the valid original offset is retained, and add a regression test for the later occurrence of a repeated hour.

src/main/java/com/ezylang/evalex/functions/datetime/DateTimeTruncateFunction.java
```diff
@@ -0,0 +1,156 @@
...
+ *   DT_TRUNCATE(dt, NULL, "America/Chicago") -&gt; 2026-08-08T05:00:00Z (Truncated to Day in Chicago time)
+ * </pre>
+ *
+ * *
+ *
+ * @author oswaldo.bapvic.jr
+ * @since 3.8.0
+ */
+@FunctionParameter(name = "dateTime")
+@FunctionParameter(name = "optionalParams", isVarArg = true) // timeUnit and zoneId
```
uklimaschewski, line 81: This unbounded vararg lets the parser accept any number of optional arguments, but `evaluate()` reads only the first two and silently ignores the rest. Please override `validatePreEvaluation()` and reject more than three total parameters with an `EvaluationException`, with a corresponding test.

src/main/java/com/ezylang/evalex/functions/datetime/DateTimeTruncateFunction.java
```diff
@@ -0,0 +1,156 @@
...
+ * {@code DT_TRUNCATE(dateTime [, timeUnit [, zoneId]])}
+ *
+ * <h3>Function Parameters</h3>
+ *
+ * <ul>
+ *   <li><b>dateTime</b> - the date-time value to truncate.
+ *   <li><b>timeUnit</b> - <i>(Optional)</i> a string containing a valid case-sensitive formatting
+ *       character. Defaults to {@code 'd'} (Day) if omitted or set to {@code NULL}:
+ *       <ul>
+ *         <li>{@code 'y'} or {@code 'u'} - Truncates to the start of the <b>Year</b>.
```
uklimaschewski: The JavaDoc advertises `u` as a supported year token, but `JAVA_FORMAT_MAPPING` does not contain it and the function rejects it. Please either add the mapping and tests or remove `u` from the documented contract.

docs/references/functions.md
```diff
@@ -2479,3 +2480,46 @@ Use this when you need a reference to today’s date without the current time co
...
+## DT_TRUNCATE
+
+*Since: 3.8.0*
+
+The `DT_TRUNCATE` function truncates a date-time value to a specific time unit, resetting smaller units to their start value, allowing developers to normalize timestamps for groupings, reports, or interval calculations. You can optionally supply a specific time unit and a target time zone. If omitted, the day ("d") unit and the system’s configured default time zone are used.
+
+### Syntax
+
+```
+DT_TRUNCATE(value [, timeUnit [, zoneId]]
```
uklimaschewski: Please close the syntax example with the final `)` and fix the missing closing backtick around the `"s"` token in the parameter table below.

uklimaschewski (changes requested):
Thanks for the contribution. The overall direction and the new DT_TRUNCATE API are useful, and the current build succeeds. Before this can be merged, please address the DST-overlap correctness issue, reject surplus parameters instead of silently ignoring them, and correct the documentation inconsistencies noted inline. Please add regression coverage for the later occurrence of a repeated DST hour, the seconds unit, and too many parameters. The 15 Sonar findings are minor S1170 warnings on existing ExpressionConfiguration fields and are not the reason for this change request. Once the corrections are pushed and CI passes, I’ll be happy to re-review.

src/main/java/com/ezylang/evalex/functions/datetime/DateTimeTruncateFunction.java
oswaldobapvicjr: Removing 'u'

src/main/java/com/ezylang/evalex/functions/datetime/DateTimeTruncateFunction.java
oswaldobapvicjr: Done

src/main/java/com/ezylang/evalex/functions/datetime/DateTimeTruncateFunction.java
oswaldobapvicjr: Good catch. Fixed and tested with new junit.

oswaldobapvicjr:
@uklimaschewski, comments addressed.

uklimaschewski (approved):
All requested corrections have been addressed. The DST-overlap behavior, parameter-count validation, documentation, and regression coverage have been re-reviewed. CI passes and SonarCloud reports no new issues. Approved.

